PR #22, opened 2020-03-30
Normalize function name to eliminate underscores

pdyraga:
go-ethereum ABI generator [drops underscores](https://github.com/ethereum/go-ethereum/blob/v1.9.12/accounts/abi/argument.go#L371) and transforms function name to camel case. Instead of `__beaconCallback`, we get `beaconCallback`. Instead of `open_keep`, we get `openKeep`. Here we introduce the same behaviour to our contract code generators.


Shadowfiend:
Don't want to merge here until we have a working build using this in keep-core or keep-ecdsa if that seems reasonable @pdyraga ?

pdyraga:
Yep, it works as expected - two PRs referenced above use the code from this PR.

Shadowfiend:
Rockin', let's do it.

PR #23, opened 2020-04-02
Concurrent execution of the event handlers

nkuba:
In this PR we allow concurrent execution of event handlers for a given subscription. Currently, when a new event comes it blocks the execution of handlers for any subsequent events.

### DIFF
--- a/tools/generators/ethereum/contract_events.go.tmpl
+++ b/tools/generators/ethereum/contract_events.go.tmpl
@@ -101,7 +101,7 @@ func ({{$contract.ShortVar}} *{{$contract.Class}}) subscribe{{$event.CapsName}}(
 					subscriptionMutex.Unlock()
 					return
 				}
-				success(
+				go success(
                     {{$event.ParamExtractors}}
 				)
 				subscriptionMutex.Unlock()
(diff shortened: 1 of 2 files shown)


pdyraga:
I am not sure about this change but would love to hear your arguments and opinion of @Shadowfiend.

I think it makes sense to do not trigger chain event handlers asynchronously. Chain processes transactions one after another and I can imagine we (or anyone else using this library) may want to process events in order as they appeared on-chain. 

Although, in my opinion, we shouldn't fire goroutines for event handler in the code of generated contracts I do recognize the problem and I think we should fire those goroutines inside the callback function, just like we do in `keep-core`. One example [here](https://github.com/keep-network/keep-core/blob/master/pkg/beacon/beacon.go#L75), another example [here](https://github.com/keep-network/keep-core/blob/master/pkg/beacon/beacon.go#L114).

Basically, the client code should decide if it cares about the order of events or not and trigger time-consuming handlers asynchronously if that's needed.

nkuba:
After pushing this PR I had similar second thoughts. The longer I test the client the more I doubt we need this.

pdyraga:
It's perfectly fine to fire callbacks in goroutines if needed, I just don't think we should do it by default. Do we want to close this PR without merging?

PR #24, opened 2020-04-07
Adding gas estimate function

dimpar:
Refs https://github.com/keep-network/keep-core/issues/1479

In this PR we add the gas estimation functionality. This will be used by Keep applications to retrieve the estimation for a specific transaction.

### DIFF
--- a/pkg/chain/ethereum/ethutil/ethutil.go
+++ b/pkg/chain/ethereum/ethutil/ethutil.go
@@ -136,6 +136,37 @@ func CallAtBlock(
 	return nil
 }
 
+// EstimateGas tries to estimate the gas needed to execute a specific transaction based on
+// the current pending state of the backend blockchain. There is no guarantee that this is
+// the true gas limit requirement as other transactions may be added or removed by miners,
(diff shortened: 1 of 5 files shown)


pkg/chain/ethereum/ethutil/ethutil.go
```diff
@@ -136,6 +136,37 @@ func CallAtBlock(
...
+) (uint64, error) {
+	input, err := contractABI.Pack(method, parameters...)
+	if err != nil {
+		return 0, err
+	}
+
+	msg := ethereum.CallMsg{
+		From: from,
+		To:   &to,
+		Data: input,
```
pdyraga, line 159: We are not passing `Value` to `CallMsg` so the estimation may not be correct for `payable` functions. I think we need to generate a variation of estimate function just for payable functions.

It's not a top priority because the problem we are having in https://github.com/keep-network/keep-core/issues/1479 is solved with the code here. I'll open a separate issue to address payable functions.

pdyraga: https://github.com/keep-network/keep-common/issues/26

PR #25, opened 2020-04-08
Checking if a storage dir is readable and writable

dimpar:
Refs https://github.com/keep-network/keep-ecdsa/issues/336

In this PR we need to check if the storage dir specified in the config .toml file under `[Storage]:DataDir` has the right read&write permission.

### DIFF
--- a/pkg/persistence/disk_persistence.go
+++ b/pkg/persistence/disk_persistence.go
@@ -12,14 +12,21 @@ const (
 )
 
 // NewDiskHandle creates on-disk data persistence handle
-func NewDiskHandle(path string) Handle {
-	err := createDir(path, currentDir)
+func NewDiskHandle(path string) (Handle, error) {
+	err := checkStoragePermission(path)
+	if err != nil {
+		return nil, err
+	}
+
+	err = createDir(path, currentDir)
 	if err != nil {
 		logger.Errorf(
 			"failed while creating directory [%v]: [%v]",
 			currentDir,
 			err,
 		)
+
+		return nil, err
 	}
 
 	err = createDir(path, archiveDir)
@@ -29,11 +36,13 @@ func NewDiskHandle(path string) Handle {
 			archiveDir,
 			err,
 		)
+
+		return nil, err
 	}
 
 	return &diskPersistence{
 		dataDir: path,
-	}
+	}, nil
 }
 
 type diskPersistence struct {
@@ -63,6 +72,22 @@ func (ds *diskPersistence) Archive(directory string) error {
 
 func (ds *diskPersistence) getStorageCurrentDirPath() string {
 	return fmt.Sprintf("%s/%s", ds.dataDir, currentDir)
+}
+
+func checkStoragePermission(dirBasePath string) error {
+	_, err := ioutil.ReadDir(dirBasePath)
+	if err != nil {
+		return fmt.Errorf("cannot read from the storage directory: [%v]", err)
+	}
+
+	tempDir, err := ioutil.TempDir(dirBasePath, "temp")
+	if err != nil {
+		return fmt.Errorf("cannot write to the storage directory: [%v]", err)
+	}
+
+	defer os.RemoveAll(tempDir)
+
+	return nil
 }
 
 func createDir(dirBasePath, newDirName string) error {
--- a/pkg/persistence/disk_persistence.go
+++ b/pkg/persistence/disk_persistence.go
@@ -80,12 +80,12 @@ func checkStoragePermission(dirBasePath string) error {
 		return fmt.Errorf("cannot read from the storage directory: [%v]", err)
 	}
 
-	tempDir, err := ioutil.TempDir(dirBasePath, "temp")
+	tempFile, err := ioutil.TempFile(dirBasePath, "write-test.*.tmp")
 	if err != nil {
 		return fmt.Errorf("cannot write to the storage directory: [%v]", err)
 	}
 
-	defer os.RemoveAll(tempDir)
+	defer os.RemoveAll(tempFile.Name())
 
 	return nil
 }
--- a/pkg/persistence/disk_persistence.go
+++ b/pkg/persistence/disk_persistence.go
@@ -20,23 +20,11 @@ func NewDiskHandle(path string) (Handle, error) {
 
 	err = createDir(path, currentDir)
 	if err != nil {
-		logger.Errorf(
-			"failed while creating directory [%v]: [%v]",
-			currentDir,
-			err,
-		)
-
 		return nil, err
 	}
 
 	err = createDir(path, archiveDir)
 	if err != nil {
-		logger.Errorf(
-			"failed while creating directory [%v]: [%v]",
-			archiveDir,
-			err,
-		)
-
 		return nil, err
 	}
 
--- a/pkg/persistence/disk_persistence_test.go
+++ b/pkg/persistence/disk_persistence_test.go
@@ -61,14 +61,14 @@ func TestDiskPersistence_StoragePermission(t *testing.T) {
 	defer os.RemoveAll(tempDir)
 
 	_, err = NewDiskHandle(tempDir)
-	if !strings.Contains(err.Error(), errExpectedRead.Error()) {
+	if err == nil || !strings.Contains(err.Error(), errExpectedRead.Error()) {
 		t.Fatalf("error on read was supposed to be returned")
 	}
 
 	os.Chmod(tempDir, 0444) // dr--r--r
 
 	_, err = NewDiskHandle(tempDir)
-	if !strings.Contains(err.Error(), errExpectedWrite.Error()) {
+	if err == nil || !strings.Contains(err.Error(), errExpectedWrite.Error()) {
 		t.Fatalf("error on write was supposed to be returned")
 	}
 }
--- a/pkg/persistence/disk_persistence_test.go
+++ b/pkg/persistence/disk_persistence_test.go
@@ -3,10 +3,11 @@ package persistence
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"strings"
 	"sync"
 	"testing"
-	"io/ioutil"
 )
 
 var (
@@ -34,7 +35,7 @@ func TestMain(m *testing.M) {
 }
 
 func TestDiskPersistence_Save(t *testing.T) {
-	diskPersistence := NewDiskHandle(dataDir)
+	diskPersistence, _ := NewDiskHandle(dataDir)
 	bytesToTest := []byte{115, 111, 109, 101, 10}
 
 	diskPersistence.Save(bytesToTest, dirName1, fileName11)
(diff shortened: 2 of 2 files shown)


dimpar:
Tested by changing permissions of the `DataDir` storage file.
ex. `chmod 000 data_storage`

pkg/persistence/disk_persistence.go
```diff
@@ -65,6 +74,22 @@ func (ds *diskPersistence) getStorageCurrentDirPath() string {
 	return fmt.Sprintf("%s/%s", ds.dataDir, currentDir)
 }
 
+func checkStoragePermission(dirBasePath string) error {
+	_, err := ioutil.ReadDir(dirBasePath)
+	if err != nil {
+		return fmt.Errorf("cannot read from the storage directory: [%v]", err)
+	}
+
+	tempDir, err := ioutil.TempDir(dirBasePath, "temp")
```
pdyraga, line 83: Maybe a file, I think it's just a little less risky because it's clear there is nothing more than this file. In case of a directory, one may think we'll write some temporary data inside while the client runs.

`ioutil.TempFile(dirBasePath, "write-test.*.tmp")` ?

pkg/persistence/disk_persistence_test.go
```diff
@@ -46,8 +47,40 @@ func TestDiskPersistence_Save(t *testing.T) {
...
+
+	_, err = NewDiskHandle(tempDir)
+	if err == nil {
+		t.Fatalf("error was supposed to be returned")
+	}
+	
+	expectedReadErr := "cannot read from the storage directory"
+	if !strings.Contains(err.Error(), expectedReadErr) {
+		t.Fatalf("error on read was supposed to be returned")
+	}
```
pdyraga, line 67: It can be simplified by creating expected error with `fmt.Errorf` and comparing later using `reflect.DeepEqual`

pkg/persistence/disk_persistence_test.go
```diff
@@ -46,8 +47,40 @@ func TestDiskPersistence_Save(t *testing.T) {
...
+	err := os.Mkdir(tempDir, 000) // d---------
+	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
+		t.Fatalf("dir [%+v] was supposed to be created", tempDir)
+	}
+	defer os.RemoveAll(tempDir)
+
+	_, err = NewDiskHandle(tempDir)
+	if err == nil {
+		t.Fatalf("error was supposed to be returned")
+	}
```
lukasz-zimnoch, line 62: If you decide to use `reflect.DeepEqual` as Piotr suggested, I think this nil check will be redundant.

pkg/persistence/disk_persistence_test.go
```diff
@@ -46,8 +47,40 @@ func TestDiskPersistence_Save(t *testing.T) {
...
+	if !strings.Contains(err.Error(), expectedReadErr) {
+		t.Fatalf("error on read was supposed to be returned")
+	}
+	
+	os.Chmod(tempDir, 0444) // dr--r--r
+	
+	_, err = NewDiskHandle(tempDir)
+	if err == nil {
+		t.Fatalf("error was supposed to be returned")
+	}
```
lukasz-zimnoch, line 74: Same here regarding the nil check.

pkg/persistence/disk_persistence_test.go
```diff
@@ -46,8 +51,30 @@ func TestDiskPersistence_Save(t *testing.T) {
...
+	tempDir := "./data_storage"
+
+	err := os.Mkdir(tempDir, 0000) // d---------
+	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
+		t.Fatalf("dir [%+v] was supposed to be created", tempDir)
+	}
+	defer os.RemoveAll(tempDir)
+
+	_, err = NewDiskHandle(tempDir)
+	if !strings.Contains(err.Error(), errExpectedRead.Error()) {
```
lukasz-zimnoch, line 64: Please use `reflect.DeepEqual` directly on `err` and `errExpectedRead` because now `err` still can be `nil`.

pkg/persistence/disk_persistence_test.go
```diff
@@ -46,8 +51,30 @@ func TestDiskPersistence_Save(t *testing.T) {
...
+		t.Fatalf("error on read was supposed to be returned")
+	}
+
+	os.Chmod(tempDir, 0444) // dr--r--r
+
+	_, err = NewDiskHandle(tempDir)
+	if !strings.Contains(err.Error(), errExpectedWrite.Error()) {
```
lukasz-zimnoch, line 71: Same here.

pkg/persistence/disk_persistence_test.go
```diff
@@ -24,6 +26,9 @@ var (
 
 	pathToCurrent = fmt.Sprintf("%s/%s", dataDir, dirCurrent)
 	pathToArchive = fmt.Sprintf("%s/%s", dataDir, dirArchive)
+
+	errExpectedRead  = errors.New("cannot read from the storage directory: ")
+	errExpectedWrite = errors.New("cannot write to the storage directory: ")
```
lukasz-zimnoch, line 31: Why not `fmt.Errorf`?

pkg/persistence/disk_persistence_test.go
dimpar: Based on your comment I started refactoring using `reflect.DeepEqual`, but then I noticed that Go lint actually doesn't recommend using `DeepEqual` for errors comparing. Currently it underlines it yellow (other files as well where we use DeepEqual)  I found [this](https://github.com/golang/go/issues/16748) github issue and this [stackoverflow](https://stackoverflow.com/questions/39121172/how-to-compare-go-errors) discussion  and based on them refactored a bit and left comparing errors without `DeepEqual`. 
Additionally, the system error will contain the data storage folder which can be named differently and that would've required additional parsing.

pkg/persistence/disk_persistence_test.go
dimpar: actually, even without `reflect.DeepEqual` it feels redundant. Removed.

pkg/persistence/disk_persistence_test.go
dimpar: removed

pkg/persistence/disk_persistence_test.go
lukasz-zimnoch: I we don't use `reflect.DeepEqual` we can do `err == nill || !strings.Contains(err.Error(), errExpectedRead.Error())`

pkg/persistence/disk_persistence_test.go
```diff
@@ -46,8 +51,30 @@ func TestDiskPersistence_Save(t *testing.T) {
...
+
+	err := os.Mkdir(tempDir, 0000) // d---------
+	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
+		t.Fatalf("dir [%+v] was supposed to be created", tempDir)
+	}
+	defer os.RemoveAll(tempDir)
+
+	_, err = NewDiskHandle(tempDir)
+	if !strings.Contains(err.Error(), errExpectedRead.Error()) {
+		t.Fatalf("error on read was supposed to be returned")
```
lukasz-zimnoch, line 65: Maybe we can follow the convention:
```
t.Errorf(
    "unexpected error\nexpected: %v\nactual:   %v\n",
    expectedError,
    err,
)
```

pkg/persistence/disk_persistence_test.go
dimpar: yea, I'll add a check for nil, regarding `reflect.DeepEqual` see [this](https://github.com/keep-network/keep-common/pull/25#discussion_r406120240) comment

pkg/persistence/disk_persistence_test.go
dimpar: I thought about it, but the full error that is returned would look smth like:
```
cannot read from the storage directory: [open ./data_storage: permission denied]
```
which contains a folder name specific to each operator. `expectedError ` is just a part of the full message that we have control over and I thought that displaying 
```
"unexpected error\nexpected: %v\nactual:   %v\n",
    expectedError,
    err,
```
is not actually true (because we display a partial error message as expected). Idk, it's not biggy and I can change it if you think it adds value.

lukasz-zimnoch: Ok, you're right. We can leave as it is.

pkg/persistence/disk_persistence.go
```diff
@@ -65,6 +74,22 @@ func (ds *diskPersistence) getStorageCurrentDirPath() string {
...
+	}
+
+	defer os.RemoveAll(tempFile.Name())
+
+	return nil
+}
+
 func createDir(dirBasePath, newDirName string) error {
 	dirPath := fmt.Sprintf("%s/%s", dirBasePath, newDirName)
 	if _, err := os.Stat(dirPath); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
dimpar: I actually dunno. `current` and `archive` are created with `rw` permissions. If later they do not have `rw` permission, it means that someone on purpose had to either change the permissions or remove folders and create new ones without `r` or `w` permission. 

I can certainly add this check if you think it is needed.

pkg/persistence/disk_persistence.go
```diff
@@ -29,11 +36,13 @@ func NewDiskHandle(path string) Handle {
 			archiveDir,
 			err,
 		)
+
+		return nil, err
```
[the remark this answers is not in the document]
dimpar: in this case, maybe we should just `return nil, err`? Otherwise, we would return sort of redundant message, because `createDir` already notifies that there were an error creating dir with the `dirPath`. What do you think?

pkg/persistence/disk_persistence.go
```diff
@@ -12,14 +12,21 @@ const (
...
+
+	err = createDir(path, currentDir)
 	if err != nil {
 		logger.Errorf(
 			"failed while creating directory [%v]: [%v]",
 			currentDir,
 			err,
 		)
+
+		return nil, err
```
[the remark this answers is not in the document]
dimpar: same as in [here](https://github.com/keep-network/keep-common/pull/25#discussion_r406144046)

pkg/persistence/disk_persistence_test.go
dimpar: I think both are fine according to [go blog on errors](https://tip.golang.org/pkg/errors/#example_New). 
And in this case, I also do not have a format specifier

pkg/persistence/disk_persistence_test.go
dimpar: added nil check

pkg/persistence/disk_persistence.go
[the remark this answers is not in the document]
nkuba: Makes sense! Please just remove `logger.Erorrf` above.

pkg/persistence/disk_persistence.go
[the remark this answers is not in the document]
nkuba: Please just remove `logger.Errorf`.

PR #27, opened 2020-04-13
Fix ethereum command function names

pdyraga:
Closes https://github.com/keep-network/keep-core/issues/1579

- In https://github.com/keep-network/keep-common/commit/8ca95ff9de561f251db07c48cdcdfb76da079670#diff-10b1a8d2fcd46a80a34cd1e5bdf789c7L198 we introduced `toCamelCase` which normalizes function names using the same strategy as go-ethereum. This was needed so that we can have `__beaconCallback` function in `BondedECDSAKeepFactory`.
- `toCamelCase` uppercases the first letter, no matter if the function name starts with `_` or not. For example, `getGroupMembers` becomes `GetGroupMembers`.
- In `buildMethodInfo` we create `dashedName`:
    ```
    shortVarRegexp, err = regexp.Compile("([A-Z])[^A-Z]*")

    // (...)

	dashedName := strings.ToLower(string(shortVarRegexp.ReplaceAll(
		[]byte(normalizedName),
		[]byte("-$0"),
	)))
    ```

   Having the first letter uppercase, `GetGroupMembers` becomes `-get-group-members`.

The fix is easy, we need to ensure the first letter stays lowercase during the normalization. Later, in `buildMethodInfo` we specify explicitly where do we want lowercase and uppercase with `uppercaseFirst` and `lowercaseFirst`.

### DIFF
--- a/tools/generators/ethereum/contract_parsing.go
+++ b/tools/generators/ethereum/contract_parsing.go
@@ -275,10 +275,18 @@ func buildEventInfo(eventsByName map[string]abi.Event) []eventInfo {
 }
 
 func uppercaseFirst(str string) string {
+	if len(str) == 0 {
+		return str
+	}
+
 	return strings.ToUpper(str[0:1]) + str[1:]
 }
 
 func lowercaseFirst(str string) string {
+	if len(str) == 0 {
+		return str
+	}
+
 	return strings.ToLower(str[0:1]) + str[1:]
 }
 
--- a/tools/generators/ethereum/contract_parsing.go
+++ b/tools/generators/ethereum/contract_parsing.go
@@ -95,7 +95,7 @@ func buildContractInfo(
 	payableMethods := make(map[string]struct{})
 	for _, methodPayableInfo := range payableInfo {
 		if methodPayableInfo.Payable {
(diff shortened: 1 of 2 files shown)


pdyraga:
Tested locally with `keep-core` and `keep-ecdsa`:
1. `relay genesis` and `relay request` work as expected
2. ethereum commands are generated as expected and work properly:
    ```
    COMMANDS:
         is-group-registered              Calls the constant method isGroupRegistered on the KeepRandomBeaconOperator contract.
         submitted-tickets                Calls the constant method submittedTickets on the KeepRandomBeaconOperator contract.
         gas-price-ceiling                Calls the constant method gasPriceCeiling on the KeepRandomBeaconOperator contract.
         ticket-submission-timeout        Calls the constant method ticketSubmissionTimeout on the KeepRandomBeaconOperator contract.
    (...)
    ```
    ```
    $ KEEP_ETHEREUM_PASSWORD="password" ./keep-core --config 
    configs/config.local.1.toml ethereum keep-random-beacon-operator number-of-groups
    1
    ```
3. `keep-ecdsa` compiles and smoke test works

PR #30, opened 2020-04-18
Upgraded version of ipfs/go-log to v2

nkuba:
This PR upgrades version of ipfs/go-log to v2. The new version allows having logs in `json` format.


nkuba:
I'm not sure about updating our loggers to use go-log v2 as libraries we are using across keep-core and keep-ecdsa are still on v1. This may cause inconsistencies in log output so maybe it would be better to postpone this change until the rest of the world catches up and handles v2.

PR #32, opened 2020-04-21
Allow to trigger cache sweep manually

pdyraga:
Previously, the cache was swept only on `Add` action which could be problematic because we may want to sweep the old entries earlier, especially for cases when `Add` is invoked rarely, like e.g. active keep check.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -49,7 +49,33 @@ func (tc *TimeCache) Add(item string) bool {
 		return false
 	}
 
-	// sweep old entries (those for which caching timespan has passed)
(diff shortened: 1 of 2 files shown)


pkg/cache/cache.go
```diff
@@ -49,7 +49,33 @@ func (tc *TimeCache) Add(item string) bool {
 		return false
```
nkuba, line 49: Maybe we should update the given item with a current `time.Now()` value?

nkuba: Oh yeah, that would mean we have to move the item to the front too. Not sure if it's worth the hassle.

PR #33, opened 2020-04-27
Fixed from address for contract call options

nkuba:
`from` address in call options for contract binding should be set to the address of the signer. `contractAddress` is supposed to be a destination of the call.


pdyraga:
Works as advertised. Subsequent PRs:
- https://github.com/keep-network/keep-core/pull/1708
- https://github.com/keep-network/keep-ecdsa/pull/432

PR #35, opened 2020-05-12
Local nonce manager

pdyraga:
The nonce manager is used by the generated contracts to track the current nonce and update it after each successfully submitted transaction. Tracking the nonce locally is required when transactions are submitted from multiple goroutines or when Ethereum clients are deployed behind a load balancer and their mempools are not always in sync.

One important aspect that might not be clear at first is that we need to call `IncrementNonce()` only when the transaction has been successfully submitted to the mempool. Also, the nonce manager has to be a single instance stored between all contracts.

`NonceManager` itself does not provide any synchronization. All generated contracts use the transaction lock provided by the chain and that transaction lock makes nonces manipulation safe.

`NonceManager` considers the local nonce value as no longer valid and updates it to the pending nonce value as seen by the chain after 5 seconds of inactivity in computing nonces. This is to let us recover from potential mempool crashes before the last TX has been propagated to other mempools. However, as long as we submit transactions one after another, the local nonce value is what determines the nonce used in transaction.

I was considering keeping `NonceManager` in `contracts` package so that `CurrentNonce()` and `IncrementNonce()` (especially this one) are not accessible from anywhere else than `contracts` but it does not look right to me that the chain would have to create `contracts.NonceManager` and pass it to each contract as a constructor parameter. Given that we have the error resolver in `ethutil`, `NonceManager ` fits better in `ethutil` IMO.

### DIFF
--- a/pkg/chain/ethereum/ethutil/nonce.go
+++ b/pkg/chain/ethereum/ethutil/nonce.go
@@ -2,10 +2,16 @@ package ethutil
 
 import (
 	"context"
+	"time"
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 )
+
+// The time for which the nonce value cached locally is valid. The local copy
+// is invalidated after the certain duration to let the nonce recover in case
+// the mempool crashed before propagating the last transaction sent.
+const localNonceTrustDuration = 5 * time.Second
 
 // NonceManager tracks the nonce for the account and allows to update it after
 // each successfully submitted transaction. Tracking the nonce locall is
@@ -23,9 +29,10 @@ import (
 // 4. Call IncrementNonce(),
 // 5. Release transaction lock.
 type NonceManager struct {
-	account    common.Address
-	transactor bind.ContractTransactor
-	localNonce uint64
+	account        common.Address
+	transactor     bind.ContractTransactor
+	localNonce     uint64
+	expirationDate time.Time
 }
 
 // NewNonceManager creates NonceManager instance for the provided account using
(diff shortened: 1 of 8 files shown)


pkg/chain/ethereum/ethutil/nonce.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// NonceManager tracks the nonce for the account and allows to update it after
+// each successfully submitted transaction. Tracking the nonce locall is
```
nkuba, line 11:
```suggestion
// each successfully submitted transaction. Tracking the nonce locally is
```

pdyraga: 4cefc7b

pkg/chain/ethereum/ethutil/nonce.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// After localNonceTrustDuration of inactivity (no CurrentNonce() calls),
+	// the local copy is considered as no longer up-to-date and it's always
+	// reset to the pending nonce value as seen by the chain.
+	//
+	// We do it to recover from potential mempool crashes.
+	//
+	// Keep in mind, the local copy is considered valid as long as transactions
+	// are submitted one after another.
+	nm.expirationDate = now.Add(localNonceTrustDuration)
```
nkuba, line 101: I'm trying to figure out why we do this here but not before L74? Are there any specific reasons?

nkuba: Please forget about this comment. I misread the code. We're not returning from the statements above but executing this line.

pdyraga: Also, there is a comparison at line `75`:
```
if now.Before(nm.expirationDate)
```
We do not want to update `nm.expirationDate` before we use the value in that comparison.

nkuba: Yeah, exactly.

PR #36, opened 2020-05-15
TX gas price bids

pdyraga:
See https://github.com/keep-network/keep-core/pull/1803

For every submitted transaction, we check in the predefined intervals if the transaction has been mined already and if not, we increase the gas price by 20% and try to submit again.

The interval at which we check for transaction status, as well as the maximum gas price, can be configured.

### DIFF
--- a/pkg/chain/ethereum/ethutil/mine_waiter.go
+++ b/pkg/chain/ethereum/ethutil/mine_waiter.go
@@ -10,7 +10,8 @@ import (
 )
 
 // MiningWaiter allows to block the execution until the given transaction is
-// mined.
+// mined as well as monitor the transaction and bump up the gas price in case
+// it is not mined in the given timeout.
 type MiningWaiter struct {
 	backend       bind.DeployBackend
 	checkInterval time.Duration
@@ -18,7 +19,17 @@ type MiningWaiter struct {
 }
 
 // NewMiningWaiter creates a new MiningWaiter instance for the provided
-// client backend.
+// client backend. It accepts two parameters setting up monitoring rules of the
+// transaction mining status.
+//
+// Check interval is the time given for the transaction to be mined. If the
+// transaction is not mined within that time, the gas price is increased by
+// 20% and transaction is replaced with the one with a higher gas price.
+//
+// Max gas price specifies the maximum gas price the client is willing to pay
+// for the transaction to be mined. The offered transaction gas price can not
+// be higher than this value. If the maximum allowed gas price is reached, no
+// further resubmission attempts are performed.
 func NewMiningWaiter(
 	backend bind.DeployBackend,
 	checkInterval time.Duration,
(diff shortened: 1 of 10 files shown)


pkg/chain/ethereum/ethutil/mine_waiter.go
```diff
@@ -0,0 +1,134 @@
...
+		// transaction not yet mined, add 20% to the previous gas price
+		gasPrice := transaction.GasPrice()
+		twentyPercent := new(big.Int).Div(gasPrice, big.NewInt(5))
+		gasPrice = new(big.Int).Add(gasPrice, twentyPercent)
+
+		// transaction not yet mined but we reached the maximum allowed gas
+		// price; giving up, we need to wait for the last submitted TX to be
+		// mined
+		if gasPrice.Cmp(mw.maxGasPrice) > 0 {
+			logger.Infof("reached the maximum allowed gas price; stopping resubmissions")
```
nkuba, line 116: Thoughts about changing level to warn?

pkg/chain/ethereum/ethutil/mine_waiter.go
```diff
@@ -0,0 +1,134 @@
...
+
+		// transaction not yet mined, add 20% to the previous gas price
+		gasPrice := transaction.GasPrice()
+		twentyPercent := new(big.Int).Div(gasPrice, big.NewInt(5))
+		gasPrice = new(big.Int).Add(gasPrice, twentyPercent)
+
+		// transaction not yet mined but we reached the maximum allowed gas
+		// price; giving up, we need to wait for the last submitted TX to be
+		// mined
+		if gasPrice.Cmp(mw.maxGasPrice) > 0 {
```
nkuba, line 115: If the calculated gasPrice is higher than max gas do we want to use max gas for the submission? Do it just once and return if the transaction with max gas price doesn't complete.
E.g.
max gas price = 50
previous gas price = 43
new gas price = 43 *1,2 = 51
We will return but we haven't submitted a tx with the max gas price yet.

pkg/chain/ethereum/ethutil/mine_waiter.go
pdyraga: I am not sure, nothing really bad happened, we are just at the maximum allowed price. 🤷

pkg/chain/ethereum/ethutil/mine_waiter.go
pdyraga: Good point, done!

pkg/chain/ethereum/ethutil/mine_waiter.go
```diff
@@ -0,0 +1,139 @@
...
+			return
+		}
+
+		// if we still have some margin, add 20% to the previous gas price
+		twentyPercent := new(big.Int).Div(gasPrice, big.NewInt(5))
+		gasPrice = new(big.Int).Add(gasPrice, twentyPercent)
+
+		// if we reached the maximum allowed gas price, submit one more time
+		// with the maximum
+		if gasPrice.Cmp(mw.maxGasPrice) > 0 {
```
nkuba, line 121: What if the initial transaction got gas price estimate higher than maxgasprice even before we started bumping it up?

pdyraga: See https://github.com/keep-network/keep-common/pull/36/commits/9036c7c35946f97b8e6c02cfc92bcd304b3d468a

PR #40, opened 2020-06-10
Metrics package

lukasz-zimnoch:
Added metrics package which provides tools useful for gathering and exposing system metrics for external monitoring tools.

### DIFF
--- a/pkg/metrics/gauge.go
+++ b/pkg/metrics/gauge.go
@@ -43,7 +43,11 @@ func (g *Gauge) expose() string {
 	}
 	labels := strings.Join(labelsStrings, ",")
 
-	body := fmt.Sprintf("%v{%v} %v %v", g.name, labels, g.value, g.timestamp)
+	if len(labels) > 0 {
+		labels = "{" + labels + "}"
+	}
+
+	body := fmt.Sprintf("%v%v %v %v", g.name, labels, g.value, g.timestamp)
 
 	return fmt.Sprintf("%v\n%v", typeLine, body)
 }
--- a/pkg/metrics/gauge.go
+++ b/pkg/metrics/gauge.go
@@ -1,0 +1,49 @@
+package metrics
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+)
+
+// Gauge is a metric type that represents a single numerical value that can
+// arbitrarily go up and down.
+type Gauge struct {
+	name   string
+	labels map[string]string
+
+	value     float64
+	timestamp int64
+	mutex     sync.RWMutex
+}
+
+// Set allows setting the gauge to an arbitrary value.
+func (g *Gauge) Set(value float64) {
+	g.mutex.Lock()
+	defer g.mutex.Unlock()
+
+	g.value = value
+	g.timestamp = time.Now().UnixNano() / int64(time.Millisecond)
+}
+
+// Exposes the gauge in the text-based exposition format.
+func (g *Gauge) expose() string {
+	g.mutex.RLock()
+	defer g.mutex.RUnlock()
+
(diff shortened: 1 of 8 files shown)


Shadowfiend:
Is there an existing package used for similar stuff in any of our existing dependencies? Thinking specifically of libp2p.

lukasz-zimnoch:
> Is there an existing package used for similar stuff in any of our existing dependencies? Thinking specifically of libp2p.

Unfortunately, I didn't find anything fitting our needs.

pkg/metrics/gauge.go
```diff
@@ -0,0 +1,53 @@
...
+	mutex     sync.RWMutex
+}
+
+// Set allows setting the gauge to an arbitrary value.
+func (g *Gauge) Set(value float64) {
+	g.mutex.Lock()
+	defer g.mutex.Unlock()
+
+	g.value = value
+	g.timestamp = time.Now().UnixNano() / int64(time.Millisecond)
```
pdyraga, line 27: I think it's better (safer) if we calculate it as:
```
time.Nanosecond * time.Now().UnixNano() / time.Millisecond
```

pkg/metrics/registry.go
```diff
@@ -0,0 +1,170 @@
+// Package `metrics` provides some tools useful for gathering and
+// exposing system metrics for external monitoring tools.
+//
+// Currently, this package is intended to use with Prometheus but
+// can be easily extended if needed. Also, not all Prometheus metric
+// types are implemented. The main motivation of creating a custom
+// package was a need to avoid usage of external unaudited dependencies.
```
pdyraga, line 7:
```
// The main motivation of creating a custom
// package was a need to avoid usage of external unaudited dependencies.
```
Let's :fire: it, it's not important in this context.

lukasz-zimnoch:
> There is a metrics package provided by `go-ethereum`: https://github.com/ethereum/go-ethereum/tree/master/metrics
> 
> Is it something we could not use for our needs? If so, can we summarize why for future-us?
> 
> If we did not explore this option, I think I am fine merging this PR and looking at `go-ethereum` metrics in the future.

Yeah, this PR overlaps several things from `go-ethereum`, but also provides some useful functionality tailored for `keep-*` clients. Specifically, the `observer` logic allows us to define reusable metric observation loops in one place (on `keep-core` side) and avoid polluting the codebase with metric primitives which must be done in case we use bare `go-ethereum` metric package. Also, this PR is rather minimalistic because for now, we don't need all metric types and other magic. At last, we can always incorporate useful things from `go-ethereum` in the future if needed.

pkg/metrics/gauge.go
lukasz-zimnoch: Can you specify why exactly?

pdyraga: It works because `time` package works with nanosecond precision (`Nanosecond  Duration = 1`), so doing `int64(time.Millisecond)` gives `1e6` but if that ever changes, we'll not be converting properly here. See https://stackoverflow.com/a/34860368/577812

lukasz-zimnoch: Ok thanks! But your proposition needs to covert `time.Now().UnixNano()` to `time.Duration` as well. So, maybe we should just use: `time.Duration(time.Now().UnixNano()) / time.Millisecond`?

pdyraga: Or maybe just the simplest possible way: `g.timestamp = time.Now().UnixNano() / 1e6` - we know that we need to divide by 1e+6 when going from millisecond to nanosecond, it's independent of how `time` package precision.

Let's also maybe add a comment to `timestamp int64` field that it's in milliseconds.

lukasz-zimnoch: Sure!

PR #42, opened 2020-07-13
allow modifying static labels

ElderOrb:
Refs https://github.com/keep-network/keep-core/issues/1930

required by https://github.com/keep-network/keep-core/pull/1913

### DIFF
--- a/pkg/metrics/registry.go
+++ b/pkg/metrics/registry.go
@@ -152,6 +152,33 @@ func (r *Registry) NewInfo(
 	return info, nil
 }
 
+// UpdateInfo updates existing info metric which will be exposed
+// through the metrics server. In case a metric doesn't exists, an error
+// will be returned.
+func (r *Registry) UpdateInfo(
+	name string,
+	labels []Label,
+) (*Info, error) {
+	r.metricsMutex.Lock()
+	defer r.metricsMutex.Unlock()
+
+	if _, exists := r.metrics[name]; !exists {
+		return nil, fmt.Errorf("metric [%v] doesn't exist", name)
+	}
+
+	if len(labels) == 0 {
+		return nil, fmt.Errorf("at least one label should be set")
+	}
+
+	info := &Info{
+		name:   name,
+		labels: processLabels(labels),
+	}
+
+	r.metrics[name] = info
+	return info, nil
+}
+
 func processLabels(
 	labels []Label,
 ) map[string]string {
--- a/pkg/metrics/registry.go
+++ b/pkg/metrics/registry.go
@@ -11,6 +11,7 @@
 package metrics
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -38,10 +39,16 @@ func NewLabel(name, value string) Label {
 	return Label{name, value}
 }
 
+type PeerInfo struct {
+	PeerId      string
+	PeerAddress string
+}
+
 // Registry performs all management of metrics. Specifically, it allows
 // to registering new metrics and exposing them through the metrics server.
 type Registry struct {
 	metrics      map[string]metric
+	peers        []PeerInfo
 	metricsMutex sync.RWMutex
 }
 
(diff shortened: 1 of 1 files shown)


lukasz-zimnoch:
@ElderOrb thanks for your contribution! I like the idea of exposing peers list, however, I think that combining them with metrics have some major drawbacks, here is why:

Our metrics exposed through `/metrics` endpoint should represent something quantifiable or countable. This is because the main purpose of this endpoint is to be used by external monitoring and alerting tools like Prometheus, which operates on time series data. Also, Prometheus handles every unique combination of key-value label pairs as a new time series, which can dramatically increase the amount of data ([look here](https://prometheus.io/docs/practices/naming/)). If we put a potentially large and volatile list of peers into the metric label, there is a high probability such a problem emerges quickly and hurt the monitoring systems. In fact, we have the `Info` "metric" but it's intended to present rather constant values like version or client identifier which can be helpful additions when building, for example, Grafana dashboards.

So, regarding the above, I think one of the options is to write a simple separate package called `diagnostics` or `debug` which will expose a separate endpoint `/diagnostics` or `/debug` and will allow exposing arbitrary data on demand, not just peers list. Then, we could instantiate it from `keep-core` and add the data in the form of functions that will be evaluated at the moment of request.

What do you think about such an approach? I think @pdyraga can also put some ideas here.

pdyraga:
Thanks @ElderOrb 🙌  I do agree with @lukasz-zimnoch though I think we could expose the connected peer count as a metric. This is a very good starting point and may satisfy most of the needs. If we want to expose a list of peers, we should do it with a separate endpoint, as suggested. I like `/diagnostics` proposition. I'd also suggest implementing it separately, given that I think we'll all agree on the connected peer count metric?

lukasz-zimnoch:
> Thanks @ElderOrb 🙌 I do agree with @lukasz-zimnoch though I think we could expose the connected peer count as a metric. This is a very good starting point and may satisfy most of the needs. If we want to expose a list of peers, we should do it with a separate endpoint, as suggested. I like `/diagnostics` proposition. I'd also suggest implementing it separately, given that I think we'll all agree on the connected peer count metric?

We're already exposing the connected peers count metric as `connected_peers_count`. Look [here](https://github.com/keep-network/keep-core/blob/master/pkg/metrics/metrics.go).

pdyraga:
> We're already exposing the connected peers count metric as `connected_peers_count`. Look [here](https://github.com/keep-network/keep-core/blob/master/pkg/metrics/metrics.go).

Nice! I was under the impression we expose only the connected bootstraps count.

ElderOrb:
Thanks for review! So the plan should be like this: 

1. Stop exposing peers list as the part of metrics
2. Rename 'peers' endpoint into 'diagnostics'
3. Move 'peers'-related code to 'diagnostics' package

Right?

ElderOrb:
> So, regarding the above, I think one of the options is to write a simple separate package called diagnostics or debug which will expose a separate endpoint /diagnostics or /debug and will allow exposing arbitrary data on demand, not just peers list

What other data needs to be exposed? On demand - means endpoint should accept some parameters? Something like /diagnostics?query=peers  ? 

Also, what about port? If it will be separate package, should it have separate configuration entry for diagnostics port in config file?

lukasz-zimnoch:
> Thanks for review! So the plan should be like this:
> 1.  Stop exposing peers list as the part of metrics
> 2. Rename 'peers' endpoint into 'diagnostics'
> 3. Move 'peers'-related code to 'diagnostics' package
>
> Right?

I think we have to do just a generic foundation here in `keep-commons`. Everything related to peers should live in `keep-core`. My proposition:

1. Create the `diagnostics` package where `DiagnosticsRegistry` struct will live. I think their API should expose:
    - `EnableServer(port int)` which will configure and enable the `/diagnostics` endpoint
    - `RegisterSource(name string, source func() string)` which will take care of registering new diagnostics sources
2. When `/diagnostics` endpoint receives a `GET` request, the registry iterates over all registered sources and invokes their source functions. Then, they return a response that contains a summary. The output format can be for example JSON or an arbitrarily structured text.
3. On `keep-core` side we instantiate a `DiagnosticRegistry` instance, enable the server on the port fetched from configuration and register a `peers` source.

> What other data needs to be exposed? 

Currently, it doesn't matter. As I mentioned above we should only create a possibility to register diagnostics sources in the future. 

> On demand - means endpoint should accept some parameters? Something like /diagnostics?query=peers ?
  
On demand - it means that data should be gathered right at the moment when `/diagnostics` receives a `GET` request. I think doing it as a continuous background process like metrics is pointless in our case. Regarding endpoint parameters, I think we can start simple and not handle them now. 

> Also, what about port? If it will be separate package, should it have separate configuration entry for diagnostics port in config file?

Yep, exactly. I think we should add a separate `[Diagnostics]` section in the config templates of `keep-core` and `keep-ecdsa`

ElderOrb:
@lukasz-zimnoch got it, thanks. I'll start clean in a separate branch then.

pdyraga:
It's probably best to discuss it in the new PR but why not `/diagnostics/peers`? We may want to add more diagnostics information in the future.

PR #43, opened 2020-08-04
Diagnositcs package

ElderOrb:
Added diagnostics package which provides tools useful for registering diagnostics callbacks for a named sources and exposing diagnostics information as a JSON, where named sources are keys and callback results are values. Can be used for providing peers list to 3rdparty diagnostics tools (or potentially official dashboard application)

### DIFF
--- a/pkg/diagnostics/diagnostics.go
+++ b/pkg/diagnostics/diagnostics.go
@@ -1,0 +1,79 @@
+// Package `diagnostics` provides some tools useful for gathering and
+// exposing arbitrary diagnositcs information for external monitoring tools.
+//
+// Possible usage: integration nodes list into dashboard
+package diagnostics
+
+import (
+	"encoding/json"
+	"io"
+	"net/http"
+	"strconv"
+	"sync"
+
+	"github.com/ipfs/go-log"
+)
+
+var logger = log.Logger("keep-diagnostics")
+
+// Registry performs all management of diagnostic. Specifically, it allows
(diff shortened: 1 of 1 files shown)


pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}()
+}
+
+func (r *DiagnosticsRegistry) RegisterSource(name string, source func() string) {
+	r.diagnosticsMutex.Lock()
+	defer r.diagnosticsMutex.Unlock()
+
+	r.diagnosticsSources[name] = source
+}
```
lukasz-zimnoch, line 56: Currently `RegisterSource` overrides the source in case the given name is already registered. I think it's ok but we should be explicit about it and add some docs for this method with a mention about their behavior in case of duplicates.

lukasz-zimnoch (commented):
Looks very good. Just a minor comment and I'm ready to merge.

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,79 @@
...
+		var jsonObject interface{}
+		err := json.Unmarshal([]byte(jsonString), &jsonObject)
+		if err == nil {
+			diagnostics[sourceName] = jsonObject
+		}
+	}
+
+	bytes, err := json.Marshal(diagnostics)
+	if err != nil {
+		return ""
```
pdyraga, line 75: Should we log the error?

lukasz-zimnoch (approved):

PR #46, opened 2020-08-17
Implementation of the Ethereum rate limiter

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-ecdsa/issues/524

Here we implement a rate-limiting wrapper for the Ethereum contract backend. Regarding the internals, it can limit the average number of requests per second using the token bucket algorithm and the request concurrency using a semaphore.

### DIFF
--- a/pkg/chain/ethereum/config.go
+++ b/pkg/chain/ethereum/config.go
@@ -45,6 +45,14 @@ type Config struct {
 	// allowed gas price is reached, no further resubmission attempts are
 	// performed.
 	MaxGasPrice uint64
+
+	// RequestsPerSecondLimit sets the maximum average number of requests
+	// per second which can be executed against the Ethereum node.
+	RequestsPerSecondLimit int
+
+	// ConcurrencyLimit sets the maximum number of concurrent requests which
+	// can be executed against the Ethereum node in the same time.
+	ConcurrencyLimit int
 }
 
 // ContractAddress finds a given contract's address configuration and returns it
--- a/pkg/chain/ethereum/config.go
+++ b/pkg/chain/ethereum/config.go
@@ -48,10 +48,14 @@ type Config struct {
 
 	// RequestsPerSecondLimit sets the maximum average number of requests
 	// per second which can be executed against the Ethereum node.
+	// All types of chain requests are rate-limited,
+	// including view function calls.
 	RequestsPerSecondLimit int
 
 	// ConcurrencyLimit sets the maximum number of concurrent requests which
-	// can be executed against the Ethereum node in the same time.
+	// can be executed against the Ethereum node at the same time.
+	// This limit affects all types of chain requests,
+	// including view function calls.
 	ConcurrencyLimit int
 }
 
--- a/pkg/chain/ethereum/ethutil/rate_limiter.go
+++ b/pkg/chain/ethereum/ethutil/rate_limiter.go
@@ -30,7 +30,7 @@ type RateLimiterConfig struct {
 	RequestsPerSecondLimit int
 
 	// ConcurrencyLimit sets the maximum number of concurrent requests which
-	// can be executed against the underlying contract backend in the same time.
+	// can be executed against the underlying contract backend at the same time.
 	ConcurrencyLimit int
 
 	// AcquirePermitTimeout determines how long a request can wait trying
--- a/pkg/chain/ethereum/ethutil/rate_limiter.go
+++ b/pkg/chain/ethereum/ethutil/rate_limiter.go
@@ -40,6 +40,8 @@ type RateLimiterConfig struct {
 
 // WrapRateLimiting wraps the given contract backend with rate limiting
 // capabilities with respect to the provided configuration.
+// All types of requests to the contract are rate-limited,
+// including view function calls.
 func WrapRateLimiting(
 	backend bind.ContractBackend,
 	config *RateLimiterConfig,
--- a/pkg/chain/ethereum/ethutil/rate_limiter.go
+++ b/pkg/chain/ethereum/ethutil/rate_limiter.go
@@ -1,0 +1,218 @@
+package ethutil
+
+import (
+	"context"
+	"fmt"
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"golang.org/x/sync/semaphore"
+	"golang.org/x/time/rate"
+	"math/big"
+	"time"
+)
+
+type rateLimiter struct {
+	bind.ContractBackend
+
+	limiter   *rate.Limiter
+	semaphore *semaphore.Weighted
+
+	acquirePermitTimeout time.Duration
+}
+
+// RateLimiterConfig represents the configuration of the rate limiter.
+type RateLimiterConfig struct {
+	// RequestsPerSecondLimit sets the maximum average number of requests
+	// per second. It's important to note that in short periods of time
+	// the actual average may exceed this limit slightly.
+	RequestsPerSecondLimit int
+
+	// ConcurrencyLimit sets the maximum number of concurrent requests which
+	// can be executed against the underlying contract backend in the same time.
+	ConcurrencyLimit int
+
+	// AcquirePermitTimeout determines how long a request can wait trying
+	// to acquire a permit from the rate limiter.
+	AcquirePermitTimeout time.Duration
+}
+
+// WrapRateLimiting wraps the given contract backend with rate limiting
+// capabilities with respect to the provided configuration.
+func WrapRateLimiting(
+	backend bind.ContractBackend,
+	config *RateLimiterConfig,
+) bind.ContractBackend {
+	rateLimiter := &rateLimiter{ContractBackend: backend}
+
+	if config.RequestsPerSecondLimit > 0 {
+		rateLimiter.limiter = rate.NewLimiter(
+			rate.Limit(config.RequestsPerSecondLimit),
+			1,
+		)
+	}
+
+	if config.ConcurrencyLimit > 0 {
+		rateLimiter.semaphore = semaphore.NewWeighted(
+			int64(config.ConcurrencyLimit),
+		)
+	}
+
+	if config.AcquirePermitTimeout > 0 {
+		rateLimiter.acquirePermitTimeout = config.AcquirePermitTimeout
+	} else {
+		rateLimiter.acquirePermitTimeout = 5 * time.Minute
+	}
+
+	return rateLimiter
+}
+
+func (rl *rateLimiter) acquirePermit() error {
+	ctx, cancel := context.WithTimeout(
+		context.Background(),
+		rl.acquirePermitTimeout,
+	)
+	defer cancel()
+
+	if rl.limiter != nil {
+		err := rl.limiter.Wait(ctx)
+		if err != nil {
+			return err
+		}
+	}
+
+	if rl.semaphore != nil {
+		err := rl.semaphore.Acquire(ctx, 1)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (rl *rateLimiter) releasePermit() {
+	if rl.semaphore != nil {
+		rl.semaphore.Release(1)
+	}
+}
+
+func (rl *rateLimiter) CodeAt(
+	ctx context.Context,
+	contract common.Address,
+	blockNumber *big.Int,
+) ([]byte, error) {
+	err := rl.acquirePermit()
+	if err != nil {
+		return nil, fmt.Errorf("cannot acquire rate limiter permit: [%v]", err)
+	}
+	defer rl.releasePermit()
+
+	return rl.ContractBackend.CodeAt(ctx, contract, blockNumber)
+}
+
+func (rl *rateLimiter) CallContract(
+	ctx context.Context,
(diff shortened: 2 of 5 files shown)


pkg/chain/ethereum/ethutil/rate_limiter.go
```diff
@@ -0,0 +1,218 @@
...
+	"golang.org/x/time/rate"
+	"math/big"
+	"time"
+)
+
+type rateLimiter struct {
+	bind.ContractBackend
+
+	limiter   *rate.Limiter
+	semaphore *semaphore.Weighted
```
pdyraga, line 20: I am not sure if we should have a `semaphore` and `ConcurrentLimit` in the rate limiter given [there is no concurrency in ethereum implementation](https://github.com/keep-network/keep-core/blob/master/pkg/chain/ethereum/connect.go#L46-L58):
```
// transactionMutex allows interested parties to forcibly serialize
// transaction submission.
//
// When transactions are submitted, they require a valid nonce. The nonce is
// equal to the count of transactions the account has submitted so far, and
// for a transaction to be accepted it should be monotonically greater than
// any previous submitted transaction. To do this, transaction submission
// asks the Ethereum client it is connected to for the next pending nonce,
// and uses that value for the transaction. Unfortunately, if multiple
// transactions are submitted in short order, they may all get the same
// nonce. Serializing submission ensures that each nonce is requested after
// a previous transaction has been submitted.
transactionMutex *sync.Mutex
```

Even if users set `ConcurrencyLimit` to `>1` in their configuration, they simply can't get it.

I'd leave the concurrency part to the underlying implementation and focus only on rate-limiting here.

pkg/chain/ethereum/ethutil/rate_limiter.go
```diff
@@ -0,0 +1,218 @@
...
+		)
+	}
+
+	if config.ConcurrencyLimit > 0 {
+		rateLimiter.semaphore = semaphore.NewWeighted(
+			int64(config.ConcurrencyLimit),
+		)
+	}
+
+	if config.AcquirePermitTimeout > 0 {
```
pdyraga, line 62: Shouldn't we set this value only when `RequestsPerSecondLimit` is set? `rateLimiter.acquirePemitTimeout` does not make sense without `rateLimiter.limiter`. What do you think about moving `if config.AcquirePermitTimeout > 0` into `if config.RequestsPerSecondLimit > 0` for clarity?

```
if config.RequestsPerSecondLimit > 0 {
    rateLimiter.limiter = // ...

   if config.AcquirePermitTimeout > 0 {
       // ...
    }
}
```

pkg/chain/ethereum/ethutil/rate_limiter_test.go
```diff
@@ -0,0 +1,246 @@
...
+	events []string
+	mutex  sync.Mutex
+}
+
+func (mb *mockBackend) CodeAt(
+	ctx context.Context,
+	contract common.Address,
+	blockNumber *big.Int,
+) ([]byte, error) {
+
```
pdyraga, line 177: Let's drop the empty line.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
```diff
@@ -0,0 +1,246 @@
...
+	if averageRequestsPerSecond > float64(requestsPerSecondLimit) {
+		t.Errorf(
+			"average requests per second exceeded the limit\n"+
+				"limit:  [%v]\n"+
+				"actual: [%v]",
+			requestsPerSecondLimit,
+			averageRequestsPerSecond,
+		)
+	}
```
pdyraga, line 69: This test takes now 5 seconds to complete and we are covering just `SendTransaction`. I think we want to confirm that all functions of our decorator are covered. I'd lower `requestDuration`, `requestsPerSecondLimit`, and/or `requests` and make sure all functions are covered.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
```diff
@@ -0,0 +1,246 @@
...
+			wg.Done()
+		}()
+	}
+
+	close(startSignal)
+
+	wg.Wait()
+
+	if len(errors) == 0 {
+		t.Fatalf("at least one timeout error should be present")
```
pdyraga, line 153: At least one or at least two given that `requestDuration = 250 * time.Millisecond` and `acquirePermitTimeout := 10 * time.Millisecond`?

pkg/chain/ethereum/ethutil/rate_limiter.go
```diff
@@ -0,0 +1,218 @@
...
+
+// RateLimiterConfig represents the configuration of the rate limiter.
+type RateLimiterConfig struct {
+	// RequestsPerSecondLimit sets the maximum average number of requests
+	// per second. It's important to note that in short periods of time
+	// the actual average may exceed this limit slightly.
+	RequestsPerSecondLimit int
+
+	// ConcurrencyLimit sets the maximum number of concurrent requests which
+	// can be executed against the underlying contract backend in the same time.
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, I mean the latter. I'll change it.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
lukasz-zimnoch: This actually doesn't matter. We only need to make sure the timeout works.

pkg/chain/ethereum/ethutil/rate_limiter.go
lukasz-zimnoch: Yeah, transactions have no concurrency but read-only calls have. And the concurrency of calls was the main problem we observed on Alchemy.

pkg/chain/ethereum/ethutil/rate_limiter.go
lukasz-zimnoch: That timeout is also used by the `rateLimiter.semaphore`

lukasz-zimnoch:
@pdyraga I'm ready for the next round.

pkg/chain/ethereum/config.go
```diff
@@ -45,6 +45,14 @@ type Config struct {
 	// allowed gas price is reached, no further resubmission attempts are
 	// performed.
 	MaxGasPrice uint64
+
+	// RequestsPerSecondLimit sets the maximum average number of requests
+	// per second which can be executed against the Ethereum node.
```
pdyraga, line 50: It wasn't clear to me when reviewing the code we want to rate-limit all types of requests. It makes a perfect sense but let's make it clear in the documentation:
```
// All types of chain requests are rate-limited, including view function calls.
```

pkg/chain/ethereum/config.go
```diff
@@ -45,6 +45,14 @@ type Config struct {
...
+	RequestsPerSecondLimit int
+
+	// ConcurrencyLimit sets the maximum number of concurrent requests which
+	// can be executed against the Ethereum node in the same time.
```
pdyraga, line 54: For the same reason as mentioned in the previous comment, I'd improve the documentation a little:
```
// This limit affects all types of chain requests, including view function calls.
```

pkg/chain/ethereum/ethutil/rate_limiter.go
pdyraga: Ah, it makes perfect sense. I think we should improve the documentation as it wasn't clear to me at all.

pkg/chain/ethereum/ethutil/rate_limiter.go
```diff
@@ -0,0 +1,218 @@
...
+	// can be executed against the underlying contract backend at the same time.
+	ConcurrencyLimit int
+
+	// AcquirePermitTimeout determines how long a request can wait trying
+	// to acquire a permit from the rate limiter.
+	AcquirePermitTimeout time.Duration
+}
+
+// WrapRateLimiting wraps the given contract backend with rate limiting
+// capabilities with respect to the provided configuration.
```
pdyraga, line 42: Let's be very explicit:
```
// All types of requests to the contract are rate-limited, including view function calls.
```

pkg/chain/ethereum/ethutil/rate_limiter.go
pdyraga: Ah, right, I missed it. `rl.semaphore.Acquire` is accepting `ctx` that has been created with `rl.acquirePermitTimeout`.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
```diff
@@ -0,0 +1,350 @@
...
+	"testing"
+	"time"
+)
+
+func TestRateLimiter(t *testing.T) {
+	requestsPerSecondLimit := 500
+	concurrencyLimit := 5
+	acquirePermitTimeout := time.Minute
+	requests := 500
+	requestDuration := 10 * time.Millisecond
```
pdyraga, line 20: Shouldn't we have a separate test for rate-limiting and concurrency level?

Even if I comment out `RequestsPerSecondLimit` in `RateLimiterConfig` this test is still passing because of how `concurrencyLimit` and `requestDuration` are configured. We can have at most `1000 / 10 * 5` requests per second.

pdyraga: Alternatively, I think we can try to generate random request durations in some range.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+	startSignal := make(chan struct{})
+	errorChannel := make(chan error, requests)
+	errors := make([]error, 0)
+
+	go func() {
+		for e := range errorChannel {
+			errors = append(errors, e)
+		}
+	}()
```
pdyraga, line 225: Instead of reading from the channel we can give this channel a buffer of `requests` - I think it's a little cleaner/easier to understand.

pdyraga:
https://github.com/keep-network/keep-ecdsa/issues/524 is not enough, we also need to do the same in `keep-core`.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
lukasz-zimnoch: It is buffered.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
lukasz-zimnoch: I can add some separate tests but in general this is how it works. Limiting only the `RequestsPerSecondLimit` always have an impact on concurrency and using the `ConcurrencyLimit` alone will impact the requests per second rate as well.

pdyraga: Yeah, I do understand they influence each other but just from the test perspective, I'd like to have a code to ensure they both work the way we expect it. Right now, we don't know whether rate-limiter works because concurrency limit is taking the entire hit in this test.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
pdyraga: Ah, right. I missed that `0` is passed to `errors` initialization call and not to `errorChannel` initialization call. Neverthless, having the buffered channel we can get rid of the entire
```
go func() {
    for e := range errorChannel {
        errors = append(errors, e)
    }
}()
```
goroutine and do:
```
errors := make(chan error, requests)
// (...)

close(errors)
if len(errors) == 0 {
  // (...)
}

for e := range errors {
  // (...)
}
```

By using the buffer, we are not leaving any goroutine from test dangling.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
lukasz-zimnoch: Ok I will rework those tests a bit.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
lukasz-zimnoch: Ah got your point. Indeed we can simplify this. I will change it.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
lukasz-zimnoch: Done.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
lukasz-zimnoch: Done.

pkg/chain/ethereum/ethutil/rate_limiter_test.go
```diff
@@ -0,0 +1,489 @@
...
+			close(startSignal)
+
+			wg.Wait()
+
+			duration := time.Now().Sub(startTime)
+			averageRequestsPerSecond := float64(requests) / duration.Seconds()
+
+			// The actual average can exceed the limit a little bit.
+			// Here we set the maximum acceptable deviation to 2%.
+			maxDeviation := 0.02
```
pdyraga, line 157: It's failing for me. I think setting it at 5% is acceptable:
```
--- FAIL: TestRateLimiter_RequestsPerSecondLimitOnly (9.01s)
    --- FAIL: TestRateLimiter_RequestsPerSecondLimitOnly/test_PendingNonceAt (0.97s)
        rate_limiter_test.go:160: average requests per second exceeded the limit
            limit:  [500]
            actual: [517.0314599021701]
```

PR #48, opened 2020-08-20
Add GoSec workflow

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/2007

Here we introduce the [GoSec](https://github.com/securego/gosec) scan workflow to the Github Actions pipeline and deal with all discovered problems.

Summary of `#nosec` exceptions and their motivation:

1. https://github.com/keep-network/keep-common/blob/024f151c484a5c59a0667dd10907236303746450/pkg/chain/ethereum/ethutil/ethutil.go#L40
Rule G304: File path provided as taint input
Motivation: We use this line to read a local key file. There is no user input.

2. https://github.com/keep-network/keep-common/blob/024f151c484a5c59a0667dd10907236303746450/pkg/generate/generate.go#L47
Rule G104: Audit errors not checked
Rule G307: Deferring a method which returns an error
Motivation: This line is placed in the auxiliary generator code, not in the core application. Also, the `Close` function returns only the error. It doesn't return any other values which can be a security threat when used without checking the error.

3. https://github.com/keep-network/keep-common/blob/024f151c484a5c59a0667dd10907236303746450/pkg/persistence/disk_persistence.go#L146
Rule G304: File path provided as taint input
Motivation: There is no user input.

4. https://github.com/keep-network/keep-common/blob/024f151c484a5c59a0667dd10907236303746450/tools/generators/ethereum/contract.go#L68
Rule G304: File path provided as taint input
Motivation: This line is placed in the auxiliary generator code, not in the core application. User input has to be passed to provide a path to the contract ABI.

5. https://github.com/keep-network/keep-common/blob/024f151c484a5c59a0667dd10907236303746450/tools/generators/ethereum/contract.go#L236
Rule G104: Audit errors not checked
Rule G307: Deferring a method which returns an error
Motivation: This line is placed in the auxiliary generator code, not in the core application. Also, the `Close` function returns only the error. It doesn't return any other values which can be a security threat when used without checking the error.

6. https://github.com/keep-network/keep-common/blob/024f151c484a5c59a0667dd10907236303746450/tools/generators/template/template.go#L38
Rule G304: File path provided as taint input
Motivation: This line is placed in the auxiliary generator code, not in the core application. User input has to be passed to provide a path to the template file.

### DIFF
--- a/.github/workflows/gosec.yml
+++ b/.github/workflows/gosec.yml
@@ -1,0 +1,20 @@
+name: Gosec
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  scan:
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    steps:
+      - uses: actions/checkout@v2
+      - uses: securego/gosec@master
+        with:
+          args: ./...
--- a/pkg/chain/ethereum/ethutil/ethutil.go
+++ b/pkg/chain/ethereum/ethutil/ethutil.go
@@ -35,7 +35,9 @@ func AddressFromHex(hex string) (common.Address, error) {
 
 // DecryptKeyFile reads in a key file and uses the password to decrypt it.
 func DecryptKeyFile(keyFile, password string) (*keystore.Key, error) {
-	data, err := ioutil.ReadFile(keyFile) // #nosec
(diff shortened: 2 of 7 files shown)


pdyraga:
Do you mind summarizing `#nosec` occurrences and why is it safe to do that?

pdyraga (approved):

PR #49, opened 2020-09-29
Add snapshot feature to the disk persistence layer

lukasz-zimnoch:
Implemented a new `Snapshot` method in the disk persistence layer. It allows storing a unique timestamped snapshot
file of given data in a dedicated `snapshot` directory.

### DIFF
--- a/pkg/persistence/disk_persistence.go
+++ b/pkg/persistence/disk_persistence.go
@@ -4,11 +4,14 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"sync"
+	"time"
 )
 
 const (
-	currentDir = "current"
-	archiveDir = "archive"
+	currentDir  = "current"
+	archiveDir  = "archive"
+	snapshotDir = "snapshot"
 
 	maxFileNameLength = 128
 )
@@ -30,13 +33,27 @@ func NewDiskHandle(path string) (Handle, error) {
 		return nil, err
 	}
 
+	err = createDir(path, snapshotDir)
+	if err != nil {
+		return nil, err
+	}
+
+	snapshotSuffixGenerator := func() string {
+		timestamp := time.Now().UnixNano() / int64(time.Millisecond)
+		return fmt.Sprintf(".%d", timestamp)
+	}
+
 	return &diskPersistence{
-		dataDir: path,
+		dataDir:                 path,
+		snapshotSuffixGenerator: snapshotSuffixGenerator,
 	}, nil
 }
 
 type diskPersistence struct {
 	dataDir string
+
+	snapshotMutex           sync.Mutex
+	snapshotSuffixGenerator func() string
 }
 
 func (ds *diskPersistence) Save(data []byte, dirName, fileName string) error {
@@ -63,6 +80,53 @@ func (ds *diskPersistence) Save(data []byte, dirName, fileName string) error {
 	}
 
 	return write(fmt.Sprintf("%s/%s/%s", dirPath, dirName, fileName), data)
+}
+
+func (ds *diskPersistence) Snapshot(data []byte, dirName, fileName string) error {
+	ds.snapshotMutex.Lock()
+	defer ds.snapshotMutex.Unlock()
+
+	if len(dirName) > maxFileNameLength {
+		return fmt.Errorf(
+			"the maximum directory name length of [%v] exceeded for [%v]",
+			maxFileNameLength,
+			dirName,
+		)
+	}
+
+	snapshotSuffix := ds.snapshotSuffixGenerator()
+
+	maxSnapshotFileNameLength := maxFileNameLength - len(snapshotSuffix)
+	if len(fileName) > maxSnapshotFileNameLength {
+		return fmt.Errorf(
+			"the maximum file name length of [%v] exceeded for [%v]",
+			maxSnapshotFileNameLength,
+			fileName,
+		)
+	}
+
+	dirPath := fmt.Sprintf("%s/%s", ds.dataDir, snapshotDir)
+	err := createDir(dirPath, dirName)
+	if err != nil {
+		return err
+	}
+
+	filePath := fmt.Sprintf("%s/%s/%s", dirPath, dirName, fileName+snapshotSuffix)
+
+	// very unlikely but better fail than overwrite an existing file
+	if canWrite := isNotExist(filePath); !canWrite {
+		return fmt.Errorf(
+			"could not create unique snapshot; " +
+				"snapshot name collision has been detected",
+		)
+	}
+
+	return write(filePath, data)
+}
+
+func isNotExist(filePath string) bool {
+	_, err := os.Stat(filePath)
+	return os.IsNotExist(err)
 }
 
 func (ds *diskPersistence) ReadAll() (<-chan DataDescriptor, <-chan error) {
--- a/pkg/persistence/disk_persistence_test.go
+++ b/pkg/persistence/disk_persistence_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"reflect"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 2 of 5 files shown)


pkg/persistence/disk_persistence.go
```diff
@@ -65,6 +82,53 @@ func (ds *diskPersistence) Save(data []byte, dirName, fileName string) error {
 	return write(fmt.Sprintf("%s/%s/%s", dirPath, dirName, fileName), data)
 }
 
+func (ds *diskPersistence) Snapshot(data []byte, dirName, fileName string) error {
```
Shadowfiend, line 85: Why implement this as a separate feature of disk persistence, instead of introducing it as a separate persistence handler?

My thinking here is: having it as a handler instead of something that must be explicitly called removes the possibility for any downstream code to forget to snapshot when it would be useful. My thought on what this would look like:

```go
	handle, err := persistence.NewDiskHandle(config.Storage.DataDir)
	if err != nil {
		return fmt.Errorf("failed while creating a storage disk handler: [%v]", err)
	}
	dataLogger := persistence.NewDataLogger(handle)
	persistence := persistence.NewEncryptedPersistence(
		dataLogger,
		config.Ethereum.Account.KeyFilePassword,
	)

```

Anything written to `persistence` would get written to `dataLogger` before being written to regular disk persistence. No downstream management, no nothing---if you write it to disk, it gets snapshotted. If you want a disk handle that doesn't snapshot for some reason, you can create that separately.

This also allows future modifications where, say, the data logger writes to disk but the disk handle is replaced by an `S3Handle` or vice versa.

Thoughts?

lukasz-zimnoch: I have made it as a separate feature because of the way how it will be used by the `keep-ecdsa` client. My reasoning was:

1. The client generates the signer 
2. Before submitting the public key on-chain, the client makes a signer snapshot but doesn't persist the signer in the regular storage yet. This is because the signer is still a "draft" until the public key gets submitted on-chain so the client can't put it in the regular storage.
3. Once the public key is submitted, the client registers the signer in the in-memory registry and in the regular disk storage as usual.

Anyway, I agree with the advantages of your approach and I'll try to follow it. It will probably need to change more things on the `keep-ecdsa` side but we'll see.

lukasz-zimnoch: @Shadowfiend I've tried to follow your idea and make snapshots implicitly but it seems we'll still need the possibility of doing snapshots explicitly. This is because of how our storage layer works. In the case of `keep-ecda`, if we want to snapshot the signer as soon as possible we have to do it just after the signer is generated and before the public key is published on-chain as I mentioned in the previous comment. If we make snapshots implicitly along with the regular save, we will be forced to put the signer in the `current` directory before their key validity is confirmed by the chain. Then we will need to check again and optionally remove the signer from the `current` directory (our storage api doesn't allow removing) if the `submitPublicKey` transaction failed. Also, having the signer in the `current` directory means that we will also have them in the in-memory keep registry which in turn means the client consider them as fully operating which may not be true if the `submitPublicKey` transaction has not been accepted yet. Last but not least, if we want to snapshot signatures in the future, it will be also easier to do it explicitly because we will not save them in the same way as we currently save keeps. What do you thing?

Shadowfiend: > Before submitting the public key on-chain, the client makes a signer snapshot but doesn't persist the signer in the regular storage yet. This is because the signer is still a "draft" until the public key gets submitted on-chain so the client can't put it in the regular storage.

This is interesting. I don't think handling new write points was an explicit goal in my head, to be honest—it was just to make sure that any write that was made had a separate record of it… I think this handles the immediate need well enough that we should move it through. I think the next step is going to be an RFC to better define what we're looking for, because that seems to be a step we missed here and led to this crossing of wires.

pkg/persistence/disk_persistence.go
```diff
@@ -65,6 +82,53 @@ func (ds *diskPersistence) Save(data []byte, dirName, fileName string) error {
...
+	dirPath := fmt.Sprintf("%s/%s", ds.dataDir, snapshotDir)
+	err := createDir(dirPath, dirName)
+	if err != nil {
+		return err
+	}
+
+	filePath := fmt.Sprintf("%s/%s/%s", dirPath, dirName, fileName+snapshotSuffix)
+
+	// very unlikely but better fail than overwrite an existing file
+	if canWrite := isNotExist(filePath); !canWrite {
```
Shadowfiend, line 117: This `if` statement feels like it's giving us more code just to invert a condition that can already be articulated. Why not `if fileAlreadyExists(filePath) { ...`?

pkg/persistence/disk_persistence_test.go
```diff
@@ -110,6 +114,176 @@ func TestDiskPersistence_RefuseSave(t *testing.T) {
 	cleanup()
 }
 
+func TestDiskPersistence_Snapshot(t *testing.T) {
+	diskHandle, _ := NewDiskHandle(dataDir)
+	bytesToTest := []byte{115, 111, 109, 101, 10}
+
+	counter := 0
+	diskHandle.(*diskPersistence).snapshotSuffixGenerator = func() string {
```
Shadowfiend, line 122: I don't love how we've made the suffix generator mutable here, though I see the use for testing. Wish we had a better way to do this, especially the collision test later on.

pkg/persistence/encrypted_persistence_test.go
```diff
@@ -85,13 +85,22 @@ func TestSaveReadAndDecryptData(t *testing.T) {
 	}
 }
 
+func TestEncryptedSnapshot(t *testing.T) {
+
```
Shadowfiend, line 89: ? :grimacing:

Shadowfiend (approved):
Left a few notes and questions, but I think this is good to merge. Let's do it.

pkg/persistence/disk_persistence.go
pdyraga: We may want to revisit the moment of writing shares to `current`. I don't think the current approach in persistence layer is wrong as it could protect us against overwriting key shares if not the chain reorganization but if we do a wider refactor and do append-only to `current`, reading only the first stored membership on client start, we can have all of it without `snapshots`.

On the other hand, the solution presented here is a little more complex but gives us better flexibility.

pkg/persistence/disk_persistence.go
```diff
@@ -30,13 +33,27 @@ func NewDiskHandle(path string) (Handle, error) {
 		return nil, err
 	}
 
+	err = createDir(path, snapshotDir)
```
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed in the follow-up PR: #50

pkg/persistence/disk_persistence.go
```diff
@@ -65,6 +82,53 @@ func (ds *diskPersistence) Save(data []byte, dirName, fileName string) error {
 	return write(fmt.Sprintf("%s/%s/%s", dirPath, dirName, fileName), data)
 }
 
+func (ds *diskPersistence) Snapshot(data []byte, dirName, fileName string) error {
+	ds.snapshotMutex.Lock()
+	defer ds.snapshotMutex.Unlock()
```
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed in the follow-up PR: #50

pkg/persistence/disk_persistence.go
lukasz-zimnoch: Addressed in the follow-up PR: #50 

I've got rid of the `canWrite` variable and renamed the `isNotExist` method a bit. But, I think we should stick to the current condition format because we can save the snapshot only if we have the `oserror.ErrNotExist` error returned from the check. This is easier to check than checking file existence because in case of an existing file the `os.Stat` method can return not just a nil-error but also several other errors types, for example `oserror.ErrPermission`.

pkg/persistence/disk_persistence_test.go
lukasz-zimnoch: Yeah, agree. But a kind of compromise was needed here.

pkg/persistence/encrypted_persistence_test.go
lukasz-zimnoch: 🤦. Removed in the follow-up PR: #50

pkg/persistence/disk_persistence.go
pdyraga: Heh, I remember it took me a while to understand how `os.Stat` when I was first reviewing this code.

PR #50, opened 2020-10-02
Improvements of the snapshot feature

lukasz-zimnoch:
Follow-up PR which addresses the remaining non-blocking comments of #49.

### DIFF
--- a/pkg/persistence/disk_persistence.go
+++ b/pkg/persistence/disk_persistence.go
@@ -102,6 +102,9 @@ func (ds *diskPersistence) Snapshot(data []byte, dirName, fileName string) error
 		)
 	}
 
+	ds.snapshotMutex.Lock()
+	defer ds.snapshotMutex.Unlock()
+
 	dirPath := fmt.Sprintf("%s/%s", ds.dataDir, snapshotDir)
 	err := ensureDirectoryExists(dirPath, dirName)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/persistence/disk_persistence.go
```diff
@@ -106,15 +103,18 @@ func (ds *diskPersistence) Snapshot(data []byte, dirName, fileName string) error
...
 	dirPath := fmt.Sprintf("%s/%s", ds.dataDir, snapshotDir)
-	err := createDir(dirPath, dirName)
+	err := ensureDirectoryExists(dirPath, dirName)
 	if err != nil {
 		return err
 	}
 
 	filePath := fmt.Sprintf("%s/%s/%s", dirPath, dirName, fileName+snapshotSuffix)
 
+	ds.snapshotMutex.Lock()
```
pdyraga, line 113: Shouldn't `ensureDirectoryExist` be included in the critical section guarded by this mutex? There is a possibility we can have two concurrent goroutines continuing because the directory does not exist and one of them will fail later on `os.Mkdir`.

pdyraga (commented):
Left one comment. Ideally, I'd like to merge it before testing https://github.com/keep-network/keep-ecdsa/pull/564.

pkg/persistence/disk_persistence.go
lukasz-zimnoch: Good point! Seems it should.

PR #52, opened 2020-10-15
Trim underscore prefix while generating event params

lukasz-zimnoch:

Underscore prefixes should be trimmed while capitalizing and uncapitalizing the first letter of the event params. This change
is required because `abigen` does the same while generating the underlying contract binding so names must be aligned to
avoid compilation errors.

### DIFF
--- a/tools/generators/ethereum/contract_parsing.go
+++ b/tools/generators/ethereum/contract_parsing.go
@@ -278,6 +278,9 @@ func uppercaseFirst(str string) string {
 	if len(str) == 0 {
 		return str
 	}
+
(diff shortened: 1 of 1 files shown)


pdyraga:
> abigen does the same while generating the underlying contract binding 

Can you please provide a link where it's done by `abigen`?

lukasz-zimnoch:
> > abigen does the same while generating the underlying contract binding
> 
> Can you please provide a link where it's done by `abigen`?

Here is a PR: https://github.com/ethereum/go-ethereum/pull/15766

pdyraga (approved):

PR #53, opened 2020-10-20
Introduce optional command generation

lukasz-zimnoch:

Added the possibility to omit the command output file argument and make the generation of the command files optional.

### DIFF
--- a/tools/generators/ethereum/contract.go
+++ b/tools/generators/ethereum/contract.go
@@ -49,9 +49,11 @@ func main() {
 
 	flag.Parse()
 
-	if flag.NArg() != 3 {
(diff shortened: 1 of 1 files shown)


pdyraga (approved):

pdyraga:
Double-checked by pointing `keep-ecdsa` dependency to `keep-common` at 19c123c and generating contracts. No problems observed.

pdyraga:
More context on why we are doing it. In https://github.com/keep-network/tbtc/pull/739 we want to generate Go bindings for `tbtc` contracts but generating commands requires adding a dependency from `tbtc` to `keep-core`. We want to avoid that and given that `tbtc.js` provides a CLI to interact with `tbtc` contracts, we decided to making command generation optional.

PR #55, opened 2020-11-03
Introduce custom Ethereum client interface

lukasz-zimnoch:

Introduced custom `EthereumClient` interface which wraps several `go-ethereum` interfaces and use it as backend in the client wrappers. The main motivation is to improve wrappers usability. For example, logging and rate-limiting clients can now be used by the block counter and mining waiter tools.

### DIFF
--- a/pkg/chain/ethereum/ethutil/ethutil.go
+++ b/pkg/chain/ethereum/ethutil/ethutil.go
@@ -21,6 +21,15 @@ import (
 
 var logger = log.Logger("keep-ethutil")
 
+// EthereumClient wraps the core `bind.ContractBackend` interface with
+// some other interfaces allowing to expose additional methods provided
+// by client implementations.
+type EthereumClient interface {
+	bind.ContractBackend
+	ethereum.ChainReader
+	ethereum.TransactionReader
+}
+
 // AddressFromHex converts the passed string to a common.Address and returns it,
 // unless it is not a valid address, in which case it returns an error. Compare
 // to common.HexToAddress, which assumes the address is valid and does not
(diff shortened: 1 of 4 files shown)


pkg/chain/ethereum/ethutil/ethutil.go
```diff
@@ -21,6 +21,15 @@ import (
 
 var logger = log.Logger("keep-ethutil")
 
+// EthereumClient wraps the core `bind.ContractBackend` interface with
+// some other interfaces allowing to expose additional methods provided
+// by client implementations.
+type EthereumClient interface {
+	bind.ContractBackend
+	ethereum.ChainReader
+	ethereum.TransactionReader
```
pdyraga, line 30: Do we use `TransactionByHash` or `TransactionReceipt` from `ethereum.TransactionReader` anywhere?

lukasz-zimnoch: We use `TransactionReceipt` in the `mine_waiter.go`

pdyraga (approved):

pdyraga:
To add some more color, this change was discussed [here](https://github.com/keep-network/keep-ecdsa/pull/585#discussion_r513351032) when we realized neither `blockCounter` nor `miningWaiter` calls are rate-limited:
 
> The fact we are wrapping `bind.ContractBackend` and that this interface knows nothing about `HeaderByNumber` or functions used by `blockCounter` or `miningWaiter` is a problem here. I think it is fine to introduce our own `ethutil.EthereumClient` interface that would embed `bind.ContractBackend` `bind.DeployBackend`, [`ethereum.ChainReader`](https://github.com/ethereum/go-ethereum/blob/v1.9.23/interfaces.go), and use that interface when defining wrappers.

PR #56, opened 2020-11-12
Add chain confirmation waiter

lukasz-zimnoch:

Added the `WaitForChainConfirmation` function which provides the ability to check the chain state once the chain reaches the given block height.

### DIFF
--- a/pkg/chain/ethereum/ethutil/ethutil.go
+++ b/pkg/chain/ethereum/ethutil/ethutil.go
@@ -213,35 +213,3 @@ func (lw *loggingWrapper) EstimateGas(ctx context.Context, msg ethereum.CallMsg)
 func WrapCallLogging(logger log.EventLogger, client EthereumClient) EthereumClient {
 	return &loggingWrapper{client, logger}
 }
-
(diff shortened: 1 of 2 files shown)


pkg/chain/ethereum/ethutil/ethutil.go
```diff
@@ -213,3 +213,35 @@ func (lw *loggingWrapper) EstimateGas(ctx context.Context, msg ethereum.CallMsg)
...
+		return false, fmt.Errorf("failed to wait for block height: [%v]", err)
+	}
+
+	result, err := stateCheck()
+	if err != nil {
+		return false, fmt.Errorf("failed to get chain state confirmation: [%v]", err)
+	}
+
+	return result, nil
+}
```
pdyraga, line 247: What do you think about putting the new code in a separate file, e.g. `pkg/chain/ethereum/ethutil/confirmations.go`? Everything that is inside `ethutil.go` is Ethereum-specific and this code is not. We could even have it outside of `ethutil` package one day. (* I'd stick with `ethutil` package for now and revisit later.)

pdyraga (approved):

PR #60, opened 2020-11-23
Parse ether value from a text

nkuba:
In a configuration file, we are providing value for `MaxGasPrice` and `BalanceAlertThreshold`. If we provide it in wei we are limited to around 9.22 ether. To be more flexible and allow providing greater value especially for balance monitoring we added a possibility to provide the value as a string with three types of units: wei, Gwei and ether. We are also supporting decimals.

Few exaples of how the value can be provided:
- `BalanceAlertThreshold = 100000000000`
- `BalanceAlertThreshold = "100000000000 wei"`
- `BalanceAlertThreshold = "100 Gwei"`
- `BalanceAlertThreshold = "2 ether"`
- `BalanceAlertThreshold = "3.4 ether"`

Refs https://github.com/keep-network/keep-core/issues/2176

### DIFF
--- a/pkg/chain/ethereum/config.go
+++ b/pkg/chain/ethereum/config.go
@@ -39,12 +39,12 @@ type Config struct {
 	// time, the gas price is increased and transaction is resubmitted.
 	MiningCheckInterval int
 
-	// MaxGasPrice specifies the maximum gas price the client is
+	// MaxGasPrice specifies the maximum gas price (in wei) the client is
 	// willing to pay for the transaction to be mined. The offered transaction
 	// gas price can not be higher than the max gas price value. If the maximum
 	// allowed gas price is reached, no further resubmission attempts are
 	// performed.
-	MaxGasPrice uint64
+	MaxGasPrice *Ethers
 
 	// RequestsPerSecondLimit sets the maximum average number of requests
 	// per second which can be executed against the Ethereum node.
@@ -58,9 +58,9 @@ type Config struct {
 	// including view function calls.
 	ConcurrencyLimit int
 
-	// BalanceAlertThreshold defines a minimum value of the operator's account
+	// BalanceAlertThreshold defines a minimum value (wei) of the operator's account
 	// balance below which an alert will be triggered.
-	BalanceAlertThreshold uint64
+	BalanceAlertThreshold *Ethers
 }
 
 // ContractAddress finds a given contract's address configuration and returns it
--- a/pkg/chain/ethereum/config.go
+++ b/pkg/chain/ethereum/config.go
@@ -44,7 +44,7 @@ type Config struct {
 	// gas price can not be higher than the max gas price value. If the maximum
 	// allowed gas price is reached, no further resubmission attempts are
 	// performed.
-	MaxGasPrice *Weis
+	MaxGasPrice *Wei
 
 	// RequestsPerSecondLimit sets the maximum average number of requests
 	// per second which can be executed against the Ethereum node.
@@ -60,7 +60,7 @@ type Config struct {
 
 	// BalanceAlertThreshold defines a minimum value (wei) of the operator's account
 	// balance below which an alert will be triggered.
-	BalanceAlertThreshold *Weis
+	BalanceAlertThreshold *Wei
 }
 
 // ContractAddress finds a given contract's address configuration and returns it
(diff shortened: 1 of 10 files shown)


pkg/chain/ethereum/ethers.go
```diff
@@ -0,0 +1,69 @@
...
+
+// The most common units for Ether values.
+const (
+	Wei Unit = iota
+	Gwei
+	Ether
+)
+
+// Unit represents Ether value unit.
+type Unit int
```
pdyraga, line 29: Should `Unit` be exported given that it is used only in `UnmarshalText`?

pkg/chain/ethereum/ethers.go
```diff
@@ -0,0 +1,69 @@
...
+func (u Unit) String() string {
+	return [...]string{"wei", "Gwei", "ether"}[u]
+}
+
+// UnmarshalText is a function used to parse a value of Ethers.
+func (e *Ethers) UnmarshalText(text []byte) error {
+	re := regexp.MustCompile(`^(\d*[\.]?[\d]*)[ ]*([\w]*)$`)
+	matched := re.FindSubmatch(text)
+
+	if len(matched) < 1 {
```
pdyraga, line 40: Shouldn't it be `len(matched) < 3`? 

`FindSubmatch` at index `0` returns the leftmost match and anything to the right are the matches of subexpressions. For example:
`FindSubmatch("1 wei") = ["1 wei", "1", "wei"]`
`FindSubmatch("wei") = ["wei", "", "wei"]`
`FindSubmatch("") = ["", "", ""]`

Also, we are reading `matched[1]` and `matched[2]` later in this function so we assume they are there although it's not expressed in this check.

pkg/chain/ethereum/ethers_test.go
```diff
@@ -0,0 +1,106 @@
...
+			value:         "3,5 ether",
+			expectedError: fmt.Errorf("failed to parse value: [3,5 ether]"),
+		},
+		"invalid thousand separator": {
+			value:         "4 500 gwei",
+			expectedError: fmt.Errorf("failed to parse value: [4 500 gwei]"),
+		},
+		"invalid unit: ETH": {
+			value:         "6 ETH",
+			expectedError: fmt.Errorf("invalid unit: ETH; please use one of: wei, Gwei, ether"),
```
pdyraga, line 77: Can we add more tests for invalid values ensuring the function returns an error instead of panicking or writing down something unexpected? Some inputs worth covering as well:
- `wrrr wei`
- `wei`
- `notawei`

pkg/chain/ethereum/ethers.go
nkuba: We did `len(matched) < 1` to catch problems when text failed to be resolved with the regexp. But actually as we know what result our expression outputs we can even do `len(matched) != 3`.

pkg/chain/ethereum/ethers_test.go
```diff
@@ -0,0 +1,106 @@
+package ethereum
+
+import (
+	"fmt"
+	"math/big"
+	"reflect"
+	"testing"
+)
+
```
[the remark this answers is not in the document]
nkuba: Yeah, thanks for that! I added even more tests.

pkg/chain/ethereum/config.go
```diff
@@ -39,12 +39,12 @@ type Config struct {
...
 	MiningCheckInterval int
 
-	// MaxGasPrice specifies the maximum gas price the client is
+	// MaxGasPrice specifies the maximum gas price (in wei) the client is
 	// willing to pay for the transaction to be mined. The offered transaction
 	// gas price can not be higher than the max gas price value. If the maximum
 	// allowed gas price is reached, no further resubmission attempts are
 	// performed.
-	MaxGasPrice uint64
+	MaxGasPrice *Weis
```
pdyraga, line 47: One last comment: I think it should be `Wei`. "The gas price is 20 Gwei", "The gas price is 20000000000 wei" and not "... 20 Gweis` or "... 20000000000 weis".

nkuba: `Weis` was a workaround for a collision with a `Wei` unit. As we made the unit private we can now rename the struct to `Wei`.

pdyraga (approved):

PR #62, opened 2021-01-07
Subscription dance: backoff, logging, and test coverage

pdyraga:
Refs https://github.com/keep-network/keep-ecdsa/issues/680

This PR is the first one in the series of PRs refactoring Ethereum event subscriptions mechanism I plan to open in the next few days.

You can see it in action here: https://github.com/keep-network/keep-ecdsa/pull/663.

So far resubscription logging was implemented in contract binding templates. There were three problems with that mechanism:
1. No resubscribe backoff. Each failed subscription was retried 5 seconds after it failed. For many individual subscriptions ECDSA client is opening, a massive retry with no backoff could be interpreted as misbehavior by a third party Ethereum provider and further attempts could be completely blocked. Some operators experienced it with Alchemy.
2. All the logging was on the warning level. With the lack of backoff mentioned in the previous point, this could produce a deadly mixture - keep client could be trying to reconnect for a long time, Ethereum client could be rejecting those attempts interpreting them as DoS/misbehavior, and operator received only warnings with no single error.
3. Lack of test coverage for the resubscription mechanism. All the subscription code was placed in templates used to generate Go contract bindings and was very hard - if possible at all - to test.

Here we address all those problems.

Instead of implementing resubscriptions on our side, we wrap the code from `github.com/ethereum/go-ethereum/event` with some additional logging logic. This code is used from contract templates to keep the subscription alive. We lean on `go-ethereum` to do those resubscriptions right and we cover our logic addition in unit tests on our side.

The code from `go-ethereum` implements backoffs, increasing the delay twice until it reaches the maximum backoff time which is set in our bindings to 2 minutes.

The new code in `ethutil` wrapping `go-ethereum`'s resubscriber, allows to log an error if the subscription is dropped too often which may indicate problems with Ethereum client. The threshold is set to 15 minutes.

### DIFF
--- a/pkg/chain/ethereum/ethutil/resubscribe.go
+++ b/pkg/chain/ethereum/ethutil/resubscribe.go
@@ -29,7 +29,7 @@ import (
 // goroutine and thus are non-blocking.
 func WithResubscription(
 	backoffMax time.Duration,
-	resubscribeFn event.ResubscribeFunc,
+	subscribeFn event.ResubscribeFunc,
 	alertThreshold time.Duration,
 	thresholdViolatedFn func(time.Duration),
 	subscriptionFailedFn func(error),
@@ -44,7 +44,7 @@ func WithResubscription(
 
 		lastAttempt = now
 
-		sub, err := resubscribeFn(ctx)
+		sub, err := subscribeFn(ctx)
 		if err != nil {
 			go subscriptionFailedFn(err)
 		}
--- a/pkg/chain/ethereum/ethutil/resubscribe.go
+++ b/pkg/chain/ethereum/ethutil/resubscribe.go
@@ -1,0 +1,39 @@
+package ethutil
+
+import (
+	"context"
+	"time"
+	"github.com/ethereum/go-ethereum/event"
+)
+
+// WithResubscription wraps the resubscription function to call it repeatedly
+// to keep a subscription established. When subscription is established, it is
+// monitored and in the case of a failure, resubscription is attempted by
+// calling the function again.
+//
+// The mechanism applies backoff between calls to resubscribe function.
+// The time between calls is adapted based on the error rate, but will never
+// exceed backoffMax.
+//
+// The mechanism monitors the time elapsed between resubscription attempts and
+// if it is shorter than the specificed alertThreshold, it calls the alertFn
+// to alarm about potential problems with the stability of the subscription.
+func WithResubscription(
+	backoffMax time.Duration,
+	resubscribeFn event.ResubscribeFunc,
+	alertThreshold time.Duration,
+	alertFn func(),
+) event.Subscription {
+	lastAttempt := time.Time{}
+	wrappedResubscribeFn := func(ctx context.Context) (event.Subscription, error) {
+		now := time.Now()
+		elapsed := now.Sub(lastAttempt)
+		if elapsed < alertThreshold {
+			alertFn()
+		}
+		lastAttempt = now
+		return resubscribeFn(ctx)
+	}
+
+	return event.Resubscribe(backoffMax, wrappedResubscribeFn)
+}
--- a/pkg/chain/ethereum/ethutil/resubscribe.go
+++ b/pkg/chain/ethereum/ethutil/resubscribe.go
@@ -3,6 +3,7 @@ package ethutil
 import (
 	"context"
 	"time"
+
 	"github.com/ethereum/go-ethereum/event"
 )
 
@@ -16,45 +17,36 @@ import (
 // exceed backoffMax.
 //
 // The mechanism monitors the time elapsed between resubscription attempts and
-// if it is shorter than the specificed alertThreshold, it writes the time
-// elapsed between resubscription attempts to the thresholdViolated channel to
-// alarm about potential problems with the stability of the subscription.
+// if it is shorter than the specificed alertThreshold, it calls
+// thresholdViolatedFn passing the time elapsed between resubscription attempts.
+// This function alarms about potential problems with the stability of the
+// subscription.
 //
-// Every error returned by the wrapped subscription function, is written to
-// subscriptionFailed channel.
+// In case of an error returned by the wrapped subscription function,
+// subscriptionFailedFn is called with the underlying error.
 //
-// If the calling code is interested in reading from thresholdViolated and/or
-// subscriptionFailed channel, appropriate readers need to be set up *before*
-// WithResubscription is called.
-//
-// Writes to thresholdViolated and subscriptionFailed channels are non-blocking
-// and are not stopping resubscription attempts.
+// thresholdViolatedFn and subscriptionFailedFn calls are executed in a separate
+// goroutine and thus are non-blocking.
 func WithResubscription(
 	backoffMax time.Duration,
-	resubscribeFn event.ResubscribeFunc,	
+	resubscribeFn event.ResubscribeFunc,
 	alertThreshold time.Duration,
-	thresholdViolated chan<- time.Duration,
-	subscriptionFailed chan<- error,
+	thresholdViolatedFn func(time.Duration),
+	subscriptionFailedFn func(error),
 ) event.Subscription {
 	lastAttempt := time.Time{}
 	wrappedResubscribeFn := func(ctx context.Context) (event.Subscription, error) {
 		now := time.Now()
 		elapsed := now.Sub(lastAttempt)
 		if elapsed < alertThreshold {
-			select {
-			case thresholdViolated <- elapsed: 
-			default:
-			}
+			go thresholdViolatedFn(elapsed)
 		}
-		
+
 		lastAttempt = now
 
 		sub, err := resubscribeFn(ctx)
 		if err != nil {
-			select {
-			case subscriptionFailed <- err:
-			default:
-			}
+			go subscriptionFailedFn(err)
 		}
 		return sub, err
 	}
--- a/pkg/chain/ethereum/ethutil/resubscribe.go
+++ b/pkg/chain/ethereum/ethutil/resubscribe.go
@@ -6,33 +6,57 @@ import (
 	"github.com/ethereum/go-ethereum/event"
 )
 
-// WithResubscription wraps the resubscription function to call it repeatedly
-// to keep a subscription established. When subscription is established, it is
-// monitored and in the case of a failure, resubscription is attempted by
-// calling the function again.
+// WithResubscription wraps the subscribe function to call it repeatedly
+// to keep a subscription alive. When a subscription is established, it is
+// monitored and in the case of a failure, resubscribe is attempted by
+// calling the subscribe function again.
 //
-// The mechanism applies backoff between calls to resubscribe function.
+// The mechanism applies backoff between resubscription attempts.
 // The time between calls is adapted based on the error rate, but will never
 // exceed backoffMax.
 //
 // The mechanism monitors the time elapsed between resubscription attempts and
-// if it is shorter than the specificed alertThreshold, it calls the alertFn
-// to alarm about potential problems with the stability of the subscription.
+// if it is shorter than the specificed alertThreshold, it writes the time
+// elapsed between resubscription attempts to the thresholdViolated channel to
+// alarm about potential problems with the stability of the subscription.
+//
+// Every error returned by the wrapped subscription function, is written to
+// subscriptionFailed channel.
+//
+// If the calling code is interested in reading from thresholdViolated and/or
+// subscriptionFailed channel, appropriate readers need to be set up *before*
+// WithResubscription is called.
+//
+// Writes to thresholdViolated and subscriptionFailed channels are non-blocking
+// and are not stopping resubscription attempts.
 func WithResubscription(
 	backoffMax time.Duration,
-	resubscribeFn event.ResubscribeFunc,
+	resubscribeFn event.ResubscribeFunc,	
 	alertThreshold time.Duration,
-	alertFn func(),
+	thresholdViolated chan<- time.Duration,
+	subscriptionFailed chan<- error,
 ) event.Subscription {
 	lastAttempt := time.Time{}
 	wrappedResubscribeFn := func(ctx context.Context) (event.Subscription, error) {
 		now := time.Now()
 		elapsed := now.Sub(lastAttempt)
 		if elapsed < alertThreshold {
-			alertFn()
+			select {
+			case thresholdViolated <- elapsed: 
+			default:
+			}
 		}
+		
 		lastAttempt = now
-		return resubscribeFn(ctx)
+
+		sub, err := resubscribeFn(ctx)
+		if err != nil {
+			select {
+			case subscriptionFailed <- err:
+			default:
+			}
+		}
+		return sub, err
 	}
 
 	return event.Resubscribe(backoffMax, wrappedResubscribeFn)
--- a/pkg/chain/ethereum/ethutil/resubscribe_test.go
+++ b/pkg/chain/ethereum/ethutil/resubscribe_test.go
@@ -52,7 +52,7 @@ func TestEmitOriginalError(t *testing.T) {
 	// That failure should refer the original error.
 	err := <-subscriptionFailed
 	if err.Error() != expectedFailMessage {
-		t.Fatalf(
+		t.Errorf(
 			"unexpected subscription error message\nexpected: [%v]\nactual:   [%v]",
 			expectedFailMessage,
 			err.Error(),
@@ -99,7 +99,7 @@ func TestResubscribeAboveThreshold(t *testing.T) {
 	// in a time shorter than 150ms one after another.
 	violationCount := len(thresholdViolated)
 	if violationCount != 0 {
-		t.Fatalf(
+		t.Errorf(
 			"threshold violation reported [%v] times, expected none",
 			violationCount,
 		)
@@ -111,7 +111,7 @@ func TestResubscribeAboveThreshold(t *testing.T) {
 	// successful and had not to be retried.
 	expectedResubscriptionCalls := plannedSubscriptionFailures + 1
 	if resubscribeFnCalls != expectedResubscriptionCalls {
-		t.Fatalf(
+		t.Errorf(
 			"resubscription called [%v] times, expected [%v]",
 			resubscribeFnCalls,
 			expectedResubscriptionCalls,
@@ -121,7 +121,7 @@ func TestResubscribeAboveThreshold(t *testing.T) {
 	// Expect all subscription failures to be reported.
 	subscriptionFailCount := len(subscriptionFailed)
 	if subscriptionFailCount != plannedSubscriptionFailures {
-		t.Fatalf(
+		t.Errorf(
 			"subscription failure reported [%v] times, expected [%v]",
 			subscriptionFailCount,
 			plannedSubscriptionFailures,
@@ -171,7 +171,7 @@ func TestResubscribeBelowThreshold(t *testing.T) {
 	// resubscription attempts.
 	violationCount := len(thresholdViolated)
 	if violationCount != plannedSubscriptionFailures {
-		t.Fatalf(
+		t.Errorf(
 			"threshold violation reported [%v] times, expected [%v]",
 			violationCount,
 			plannedSubscriptionFailures,
@@ -184,7 +184,7 @@ func TestResubscribeBelowThreshold(t *testing.T) {
 	for i := 0; i < violationCount; i++ {
 		violation := <-thresholdViolated
 		if violation < elapsedBetweenFailures {
-			t.Fatalf(
+			t.Errorf(
 				"violation reported should be longer than the time elapsed "+
 					"between failures; is: [%v] and should be longer than [%v]",
 				violation,
@@ -192,7 +192,7 @@ func TestResubscribeBelowThreshold(t *testing.T) {
 			)
 		}
 		if violation > alertThreshold {
-			t.Fatalf(
+			t.Errorf(
 				"violation reported should be shorter than the alert threshold; "+
 					"; is: [%v] and should be shorter than [%v]",
 				violation,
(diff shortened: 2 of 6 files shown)


nkuba (commented):
First part of comments.

lukasz-zimnoch (changes requested):

pkg/chain/ethereum/ethutil/resubscribe_test.go
```diff
@@ -0,0 +1,280 @@
...
+		alertThreshold,
+		func(elapsed time.Duration) { thresholdViolated <- elapsed },
+		func(err error) { subscriptionFailed <- err },
+	)
+	<-subscription.Err()
+
+	// Subscription failed one time so there should be one error in the channel.
+	subscriptionFailCount := len(subscriptionFailed)
+	if subscriptionFailCount != 1 {
+		t.Fatalf(
```
nkuba, line 42: Can we replace `t.Fatalf` with `t.Errorf` in all added tests so the test reports all verification failures instead of aborting on the first one?

pkg/chain/ethereum/ethutil/resubscribe_test.go
```diff
@@ -0,0 +1,280 @@
...
+	// blocked by the lack of channel receivers on non-buffered channels.
+	expectedResubscriptionCalls := plannedSubscriptionFailures + 1
+	if resubscribeFnCalls != expectedResubscriptionCalls {
+		t.Fatalf(
+			"resubscription called [%v] times, expected [%v]",
+			resubscribeFnCalls,
+			expectedResubscriptionCalls,
+		)
+	}
+}
```
nkuba, line 280: Thoughts on adding a test where we check that if the subscription is cancelled callback functions won't be called anymore?

tools/generators/ethereum/contract_events.go.tmpl
```diff
@@ -37,119 +37,61 @@ func ({{$contract.ShortVar}} *{{$contract.Class}}) Past{{$event.CapsName}}Events
...
 
-	return subscription.NewEventSubscription(unsubscribeCallback), nil
+	sub := ethutil.WithResubscription(
+		{{$contract.ShortVar}}SubscriptionBackoffMax,
+		subscribeFn,
+		{{$contract.ShortVar}}SubscriptionAlertThreshold,
+		func(elapsed time.Duration) {
+			{{$logger}}.Errorf(
+					"subscription to event {{$event.CapsName}} had to be "+
+						"retried [%v] since the last attempt; please inspect "+
```
nkuba, line 76: Will it print fine with `%v` shouldn't it be `%s`?
```suggestion
						"retried [%s] since the last attempt; please inspect "+
```

tools/generators/ethereum/contract_events.go.tmpl
```diff
@@ -37,119 +37,61 @@ func ({{$contract.ShortVar}} *{{$contract.Class}}) Past{{$event.CapsName}}Events
...
+						"Ethereum client connectivity",
```
nkuba, line 77: Maybe we should drop `client`? Someone can connect to an ethereum client but others can use providers like Infura or Alchemy.

```suggestion
						"Ethereum connectivity",
```

tools/generators/ethereum/contract_events.go.tmpl
```diff
@@ -37,119 +37,61 @@ func ({{$contract.ShortVar}} *{{$contract.Class}}) Past{{$event.CapsName}}Events
...
+		)
 	}
 
-	return subscription.NewEventSubscription(unsubscribeCallback), nil
+	sub := ethutil.WithResubscription(
+		{{$contract.ShortVar}}SubscriptionBackoffMax,
+		subscribeFn,
+		{{$contract.ShortVar}}SubscriptionAlertThreshold,
+		func(elapsed time.Duration) {
+			{{$logger}}.Errorf(
```
nkuba, line 74: Error or Warn, that is the question.

tools/generators/ethereum/contract_events.go.tmpl
```diff
@@ -37,119 +37,61 @@ func ({{$contract.ShortVar}} *{{$contract.Class}}) Past{{$event.CapsName}}Events
...
+					"subscription to event {{$event.CapsName}} had to be "+
+						"retried [%v] since the last attempt; please inspect "+
+						"Ethereum client connectivity",
+					elapsed,
+				)
+		},
+		func(err error) {
+			{{$logger}}.Errorf(
```
nkuba, line 82: This vs L74. One of them could be Error and the other Warn. Maybe this could be a Warn and the other one Error?

pkg/chain/ethereum/ethutil/resubscribe_test.go
```diff
@@ -0,0 +1,280 @@
...
+	thresholdViolated := make(chan time.Duration, 10)
+	subscriptionFailed := make(chan error, 10)
+	subscription := WithResubscription(
+		backoffMax,
+		resubscribeFn,
+		alertThreshold,
+		func(elapsed time.Duration) { thresholdViolated <- elapsed },
+		func(err error) { subscriptionFailed <- err },
+	)
+	<-subscription.Err()
```
nkuba, line 37: How does it work? When does it resolve? Seems like resubscription is kept forever but when eventually it returns an error?

pkg/chain/ethereum/ethutil/resubscribe.go
```diff
@@ -0,0 +1,55 @@
...
+// thresholdViolatedFn and subscriptionFailedFn calls are executed in a separate
+// goroutine and thus are non-blocking.
+func WithResubscription(
+	backoffMax time.Duration,
+	resubscribeFn event.ResubscribeFunc,
+	alertThreshold time.Duration,
+	thresholdViolatedFn func(time.Duration),
+	subscriptionFailedFn func(error),
+) event.Subscription {
+	lastAttempt := time.Time{}
```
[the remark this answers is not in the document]
nkuba: Yeah, seems like my comment doesn't make sense. Sorry about that!

tools/generators/ethereum/contract.go.tmpl
```diff
@@ -25,6 +25,21 @@ import (
...
+
+    // Threshold below which event resubscription emits an error to the logs.
+    // WS connection can be dropped at any moment and event resubscription will
+    // follow. However, if WS connection for event subscription is getting
+    // dropped too often, it may indicate something is wrong with Ethereum
+    // client. This constant defines the minimum lifetime of an event
+    // subscription required before the subscription failure happens and
+    // resubscription follows so that the resubscription does not emit an error
+    // to the logs alerting about potential problems with Ethereum client.
+    {{.ShortVar}}SubscriptionAlertThreshold = 15 * time.Minute
```
Shadowfiend, line 40: 🤔 should this be per-contract?

pkg/chain/ethereum/ethutil/resubscribe.go
```diff
@@ -0,0 +1,55 @@
...
+	wrappedResubscribeFn := func(ctx context.Context) (event.Subscription, error) {
+		now := time.Now()
+		elapsed := now.Sub(lastAttempt)
+		if elapsed < alertThreshold {
```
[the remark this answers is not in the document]
pdyraga: There is no `Before` function defined for `time.Duration`. I think it's fine to use `<` in this case because both `elapsed` and `alertThreshold` are `time.Duration`.

pkg/chain/ethereum/ethutil/resubscribe.go
```diff
@@ -0,0 +1,55 @@
...
+// subscriptionFailedFn is called with the underlying error.
+//
+// thresholdViolatedFn and subscriptionFailedFn calls are executed in a separate
+// goroutine and thus are non-blocking.
+func WithResubscription(
+	backoffMax time.Duration,
+	resubscribeFn event.ResubscribeFunc,
+	alertThreshold time.Duration,
+	thresholdViolatedFn func(time.Duration),
+	subscriptionFailedFn func(error),
```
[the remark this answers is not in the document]
pdyraga: How important is this change for you? I started with something similar but then switched to shorter names to keep this code easier to read.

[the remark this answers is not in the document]
pdyraga: This function is also called if the first-ever subscription attempt fails. My reasoning was that the first attempt is *a subscription*  and even if we have to resubscribe later, resubscription attempts are also... *subscription* attempts, just repeated. That's why I named it `subscriptionFailedFn`.

In fact `resubscribeFn` could be renamed to `subscribeFn` but given that the function type from `github.com/ethereum/go-ethereum/event` is `event.ResubscribeFn` I left it as `resubscribeFn`. Happy to change it though if you think it makes sense.

pkg/chain/ethereum/ethutil/resubscribe_test.go
```diff
@@ -0,0 +1,280 @@
...
+			return nil, fmt.Errorf("wherever I go, he goes")
+		}
+		delegate := event.NewSubscription(func(unsubscribed <-chan struct{}) error {
+			return nil
+		})
+		return delegate, nil
+	}
+
+	thresholdViolated := make(chan time.Duration, 10)
+	subscriptionFailed := make(chan error, 10)
```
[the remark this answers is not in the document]
pdyraga: Added:
```
	// Using buffered channels to do not block writes.
	// There should never be a need to write more to those channels if the code
	// under the test works as expected.
```

tools/generators/ethereum/contract_events.go.tmpl
```diff
@@ -37,119 +37,61 @@ func ({{$contract.ShortVar}} *{{$contract.Class}}) Past{{$event.CapsName}}Events
...
+					"subscription to event {{$event.CapsName}} failed "+
+						"with error: [%v]; resubscription attempt will be "+
+						"performed",
+					err,
+				)
+		},
```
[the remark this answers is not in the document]
pdyraga: See eb5ee3d0578ab0adde55c7e6d732c8d7f40f5e83 please.

pkg/chain/ethereum/ethutil/resubscribe_test.go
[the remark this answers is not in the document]
pdyraga: See 44412b59cf176977a2f301cf75ed5d57380846ad please.

pkg/chain/ethereum/ethutil/resubscribe_test.go
```diff
@@ -0,0 +1,280 @@
...
+			return nil, fmt.Errorf("this is the way")
+		}
+		delegate := event.NewSubscription(func(unsubscribed <-chan struct{}) error {
+			return nil
+		})
+		return delegate, nil
+	}
+
+	thresholdViolated := make(chan time.Duration, 10)
+	subscriptionFailed := make(chan error, 10)
```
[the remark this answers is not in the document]
pdyraga: See 44412b59cf176977a2f301cf75ed5d57380846ad please.

pkg/chain/ethereum/ethutil/resubscribe_test.go
```diff
@@ -0,0 +1,280 @@
...
+			return nil, fmt.Errorf("i have spoken")
+		}
+		delegate := event.NewSubscription(func(unsubscribed <-chan struct{}) error {
+			return nil
+		})
+		return delegate, nil
+	}
+
+	thresholdViolated := make(chan time.Duration, 10)
+	subscriptionFailed := make(chan error, 10)
```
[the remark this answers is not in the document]
pdyraga: See 44412b59cf176977a2f301cf75ed5d57380846ad please.

pkg/chain/ethereum/ethutil/resubscribe.go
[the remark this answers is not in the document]
pdyraga: This was also brought up by @nkuba here: https://github.com/keep-network/keep-common/pull/62#discussion_r558288557. I will do `resubscribeFn` -> `subscribeFn` rename.

pkg/chain/ethereum/ethutil/resubscribe.go
```diff
@@ -0,0 +1,55 @@
...
+// subscription.
+//
+// In case of an error returned by the wrapped subscription function,
+// subscriptionFailedFn is called with the underlying error.
+//
+// thresholdViolatedFn and subscriptionFailedFn calls are executed in a separate
+// goroutine and thus are non-blocking.
+func WithResubscription(
+	backoffMax time.Duration,
+	resubscribeFn event.ResubscribeFunc,
```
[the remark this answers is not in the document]
pdyraga: Yes, it makes sense. We discussed it earlier here: https://github.com/keep-network/keep-common/pull/62#discussion_r565173008

tl;dr; I left it as `resubscribeFn` because of the type name but was not particularly happy about it. Since you are bringing this up, I am happy to do the rename.

pkg/chain/ethereum/ethutil/resubscribe.go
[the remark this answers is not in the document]
pdyraga: See e81b7de053718e5b4d6c67275f7528529af92854 please.

pkg/chain/ethereum/ethutil/resubscribe.go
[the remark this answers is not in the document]
pdyraga: See e81b7de053718e5b4d6c67275f7528529af92854 please.

tools/generators/ethereum/contract_events.go.tmpl
```diff
@@ -37,119 +37,61 @@ func ({{$contract.ShortVar}} *{{$contract.Class}}) Past{{$event.CapsName}}Events
 
 func ({{$contract.ShortVar}} *{{$contract.Class}}) Watch{{$event.CapsName}}(
 	success {{$contract.FullVar}}{{$event.CapsName}}Func,
-	fail func(err error) error,
 	{{$event.IndexedFilterDeclarations -}}
-) (subscription.EventSubscription, error) {
-    errorChan := make(chan error)
-    unsubscribeChan := make(chan struct{})
+) (subscription.EventSubscription) {
+	eventOccurred := make(chan *abi.{{$contract.AbiClass}}{{$event.CapsName}})
```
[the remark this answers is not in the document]
pdyraga: Agreed though this line will go away in #63 so I suggest we leave it as is here.

tools/generators/ethereum/contract_events.go.tmpl
```diff
@@ -37,119 +37,61 @@ func ({{$contract.ShortVar}} *{{$contract.Class}}) Past{{$event.CapsName}}Events
...
-	if err != nil {
-		close(eventChan)
-		return eventSubscription, fmt.Errorf(
-			"error creating watch for {{$event.CapsName}} events: [%v]",
-			err,
-		)
-	}
-
-	var subscriptionMutex = &sync.Mutex{}
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
pdyraga: Please see 7a5fcf12f04fefbab6137b71a21e483050482e70.

tools/generators/ethereum/contract_events.go.tmpl
pdyraga: Both of them should be logged on the error level. `thresholdViolatedFn` is called when resubscribe attempt had to be performed too soon. If the connection is getting broken every few minutes, it is clear something is wrong and there is a chance the client will not be able to receive events on time. This should be escalated.

tools/generators/ethereum/contract_events.go.tmpl
pdyraga: Both of them should be logged on the error level. `subscriptionFailedFn` is called when a subscribe attempt fails right away. It is fine if ws connection gets broken after several days and the subscription has to be retried. But it is very alarming if it fails right after the connect attempt.

pkg/chain/ethereum/ethutil/resubscribe_test.go
```diff
@@ -0,0 +1,280 @@
...
+	subscriptionFailCount := len(subscriptionFailed)
+	if subscriptionFailCount != 1 {
+		t.Fatalf(
+			"subscription failure reported [%v] times, expected [1]",
+			subscriptionFailCount,
+		)
+	}
+
+	// That failure should refer the original error.
+	expectedFailMessage := "wherever I go, he goes"
```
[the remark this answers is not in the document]
pdyraga: Sure. See eb2cc4af228eca465b33f4ad92ce6c1fb295368d please.

pkg/chain/ethereum/ethutil/resubscribe_test.go
pdyraga: The reason I used `Fatalf` in this case is that we should not execute the other check if we do not want to risk the test to hang. If this channel is empty, `err := <-subscriptionFailed` execution in the next line will hang. For other cases, I do agree we can use `Errorf`.

pdyraga: See 4841e64672d7215b060ca21758d9576d44f292e5 please.

pkg/chain/ethereum/ethutil/resubscribe_test.go
```diff
@@ -0,0 +1,280 @@
...
+	)
+	<-subscription.Err()
+
+	// Subscription failed one time so there should be one error in the channel.
+	subscriptionFailCount := len(subscriptionFailed)
+	if subscriptionFailCount != 1 {
+		t.Fatalf(
+			"subscription failure reported [%v] times, expected [1]",
+			subscriptionFailCount,
+		)
```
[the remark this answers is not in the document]
pdyraga: I am not sure about it, same for the other mentioned cases. For simple assertions like this one, we can fit everything in one line as there is no risk that expected/actual values comparison will be hard. And the `Errof`/`Fatalf` string looks much clearer in the code.

I would suggest that we do a single-line assertion failed message in such cases and we split to multiple lines when we e.g. compare error messages or big integers or hex byte arrays. WDYT?

tools/generators/ethereum/contract.go.tmpl
pdyraga: I think it should not be but I do not have a good place to put it at. :grimacing: Maybe `subscribe_opts.go` from `ethutil` introduced in the next PR #63 is a good candidate?

pkg/chain/ethereum/ethutil/resubscribe.go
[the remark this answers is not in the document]
lukasz-zimnoch: Sure, I'm fine with `subscribeFn`

pkg/chain/ethereum/ethutil/resubscribe_test.go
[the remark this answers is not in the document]
lukasz-zimnoch: Sure, good idea!

lukasz-zimnoch (commented):
No more comments! Leaving the rest for @nkuba

tools/generators/ethereum/contract_events.go.tmpl
nkuba: OK, so we won't log an error every time resubscription happens but only if it happens too soon?

pkg/chain/ethereum/ethutil/resubscribe_test.go
pdyraga: It does resolve here:

https://github.com/ethereum/go-ethereum/blob/2e5d14170846ae72adc47467a1129e41d6800349/event/subscription.go#L52

The function passed eventually to `event.NewSubscription` by our test subscribe function is non-blocking, there is nothing to produce and that `defer` will be called closing channel returned by `Err()`. Similar pattern is used in `go-ethereum` tests:

https://github.com/ethereum/go-ethereum/blob/2e5d14170846ae72adc47467a1129e41d6800349/event/subscription_test.go#L98

tools/generators/ethereum/contract_events.go.tmpl
pdyraga: Correct. We log error in two cases:
- resubscription had to happen too soon (threshold violated),
- subscription (resubscription) failed right after it's been requested.

Nothing is logged when the subscription fails after e.g. ten days. This is how `go-ethereum/event` `Resubscribe` approaches it. However, we modified it in https://github.com/ethereum/go-ethereum/pull/22191 and I suggest we log a warning. This is out of the scope of this PR and will require `go-ethereum` dependency bump.

tools/generators/ethereum/contract_events.go.tmpl
pdyraga: Sure, please see c777eab88a4a02e21b166c54e85a48055f2a74f9.

tools/generators/ethereum/contract_events.go.tmpl
pdyraga: It will work fine with both, I can replace `%v` with `%s` for clarity.

pdyraga: See 64cab832ccbd7ae02b6e87da48ff65bf92fd8c1f please.

pkg/chain/ethereum/ethutil/resubscribe_test.go
pdyraga: Do you think it is needed? Those functions are called from the inside of a wrapped subscribe function and I expect `go-ethereum` `Resubscribe` to be implemented right and do not call subscribe again if the given subscription was cancelled.

Even if we treat this code as a black box - which I usually prefer to do - I am not sure if it is possible to write such a test - I will unsubscribe and I can't get subscribe function to be called again, it does not make sense 🤔

pdyraga:
I updated https://github.com/keep-network/keep-ecdsa/pull/663 to point to the version with all review comments addressed, which is c6adefdc082b4ee80aa6da0075e0f4baa8531f70 - the last commit.

lukasz-zimnoch (approved):

nkuba (approved):

tools/generators/ethereum/contract.go.tmpl
nkuba: I'm going to merge this PR. Please let's handle this change in #63.

PR #65, opened 2021-02-08
Add contract prefix to events subscriptions names

lukasz-zimnoch:
Here we modify the contract generator in order to add contract prefix (short name) to events subscriptions names. This change should prevent compile errors caused by redeclared names when two contracts declare the same event name.


pdyraga:
This looks good, can we see `keep-ecdsa` PR with regenerated code to confirm that it works as expected?

lukasz-zimnoch:
It's in `keep-core`: https://github.com/keep-network/keep-core/pull/2320

pdyraga (approved):

PR #67, opened 2021-02-18
Create GH Actions workflow for build&test of Go client

michalinacienciala:
As described in RFC-18, there is a need for a refactorization of Keep
and tBTC release processes in order to reduce human involvment and make
the processes faster and less error prone. A part of this task is
migration from CircleCI jobs to GitHub Actions. This commit creates a
GitHub Action workfow for building and testing of Go client.

At this point, solution utilizes EnricoMi/publish-unit-test-result-action@v1
action for publishing of the test results, but this may be replaced in the
future by use of other action or use of problem matchers.


.circleci/config.yml
```diff
@@ -6,6 +6,8 @@ executors:
       - image: circleci/golang:1.13
 
 jobs:
+# GitHub Actions equivalent of this job (.github/workflows/client.yml)
```
nkuba, line 9: As test-go is the only workflow we've got configured in Circle CI maybe we could remove this configuration file at all.

.github/workflows/client.yml
```diff
@@ -0,0 +1,49 @@
...
+on:
+  push:
+    branches:
+      # TODO: Run on master after we're fully migrated from Circle CI
+      # - master
+      - "rfc-18/**"
+  pull_request:
+      branches:
+      # TODO: Run on all branches after we're fully migrated from Circle CI
+      - "rfc-18/**"
```
nkuba, line 13: If we decide to remove Circle CI config we can switch to the final branches setup here.
```suggestion
  push:
    branches:
      - master
  pull_request:
```

nkuba (changes requested):

.circleci/config.yml
michalinacienciala: You're right. I think we can even delete whole `.circleci` folder - thats's what I just commited.

michalinacienciala:
I think it would be good to add the passing of `Go` workflow as a required condition for merging of PRs. I don't have the privileges to do that, neither has Piotr. @Shadowfiend, would you be able to configure that?

nkuba (approved):

PR #68, opened 2021-02-25
Stabilizer Muscles: Force a stable order on method and event information

Shadowfiend:
The Go contract generator had been using the parsed contract data to
generate method and event code in an order that is currently randomized
due to the use of maps. The contract parser now forcibly sorts the
method and event slices by name to ensure a stable order in the
generated code, and thus avoid diffs for committed code unless there are
contract or code generator changes.


pdyraga (approved):
🚀

PR #69, opened 2021-02-25
ETH-like utils generator 

lukasz-zimnoch:
Here we introduce a generator of common ETH-like utils which need to have their own representation in chain-specific packages. That generator should help to avoid code duplication and will make future maintenance easier. Specifically, those tools are now managed by the generator:
- `resubscribe.go`
- ` error_resolver.go`
- `logging_wrapper.go`
- `rate_limiter.go`

### DIFF
--- a/pkg/chain/celo/celoutil/celoutil.go
+++ b/pkg/chain/celo/celoutil/celoutil.go
@@ -25,10 +25,10 @@ var logger = log.Logger("keep-celoutil")
 // errors. See ResolveError below.
 const errorABIString = "[{\"constant\":true,\"outputs\":[{\"type\":\"string\"}],\"inputs\":[{\"name\":\"message\", \"type\":\"string\"}],\"name\":\"Error\", \"type\": \"function\"}]"
 
-// CeloClient wraps the core `bind.ContractBackend` interface with
+// HostChainClient wraps the core `bind.ContractBackend` interface with
 // some other interfaces allowing to expose additional methods provided
 // by client implementations.
-type CeloClient interface {
+type HostChainClient interface {
 	bind.ContractBackend
 	celo.ChainReader
 	celo.TransactionReader
@@ -203,7 +203,7 @@ func EstimateGas(
 
 // NewBlockCounter creates a new BlockCounter instance for the provided
 // Celo client.
-func NewBlockCounter(client CeloClient) (*ethlike.BlockCounter, error) {
+func NewBlockCounter(client HostChainClient) (*ethlike.BlockCounter, error) {
 	return ethlike.CreateBlockCounter(&ethlikeAdapter{client})
 }
 
@@ -211,7 +211,7 @@ func NewBlockCounter(client CeloClient) (*ethlike.BlockCounter, error) {
 // Celo client. It accepts two parameters setting up monitoring rules
 // of the transaction mining status.
 func NewMiningWaiter(
-	client CeloClient,
+	client HostChainClient,
 	checkInterval time.Duration,
 	maxGasPrice *big.Int,
 ) *ethlike.MiningWaiter {
@@ -225,7 +225,7 @@ func NewMiningWaiter(
 // NewNonceManager creates NonceManager instance for the provided account
 // using the provided Celo client.
 func NewNonceManager(
-	client CeloClient,
+	client HostChainClient,
 	account common.Address,
 ) *ethlike.NonceManager {
 	return ethlike.NewNonceManager(
--- a/pkg/chain/celo/celoutil/celoutil.go
+++ b/pkg/chain/celo/celoutil/celoutil.go
@@ -11,6 +11,7 @@ import (
 	celoclient "github.com/celo-org/celo-blockchain/ethclient"
 	"github.com/celo-org/celo-blockchain/rpc"
 	"github.com/ipfs/go-log"
+	"github.com/keep-network/keep-common/pkg/chain/celo/celoutil/client"
 	"github.com/keep-network/keep-common/pkg/chain/ethlike"
 	"io/ioutil"
 	"math/big"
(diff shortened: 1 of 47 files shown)


pdyraga:
Let's use the `gen` directory layout for the generated code. This layout is even mentioned in the [directory structure](https://github.com/keep-network/keep-common) section of `keep-common` README.

pkg/chain/celo/celoutil/celoutil.go
```diff
@@ -19,10 +19,16 @@ import (
 
 var logger = log.Logger("keep-celoutil")
 
-// CeloClient wraps the core `bind.ContractBackend` interface with
+// ABI for errors bubbled out from revert calls. Not used directly as errors are
+// neither encoded strictly as method calls nor strictly as return values, nor
+// strictly as events, but some various bits of it are used for unpacking the
+// errors. See ResolveError below.
+const errorABIString = "[{\"constant\":true,\"outputs\":[{\"type\":\"string\"}],\"inputs\":[{\"name\":\"message\", \"type\":\"string\"}],\"name\":\"Error\", \"type\": \"function\"}]"
```
pdyraga, line 26: Do we know why is it different between Ethereum and Celo? Is it because of `go-ethereum` and `celo-blockchain` dependency versions? If we update dependencies, can we have the same `errorABIString`?

pkg/chain/celo/celoutil/celoutil.go
```diff
@@ -19,10 +19,16 @@ import (
...
+
+// HostChainClient wraps the core `bind.ContractBackend` interface with
 // some other interfaces allowing to expose additional methods provided
 // by client implementations.
-type CeloClient interface {
+type HostChainClient interface {
```
pdyraga, line 31: What do you think about naming it just `Client` (or `ChainClient`)? From the perspective of `celoutil` package, the host chain is Celo so `HostChain*` feels clumsy.

pkg/chain/celo/celoutil/celoutil.go
lukasz-zimnoch: It's because `celo-blockchain` has been forked from an older version of `go-ethereum` than we use. Update of dependencies won't change anything.

pkg/chain/celo/celoutil/celoutil.go
lukasz-zimnoch: https://github.com/keep-network/keep-common/commit/2eae82a2416dc5916848986c9a55094c9999d6e8

lukasz-zimnoch:
> Let's use the `gen` directory layout for the generated code. This layout is even mentioned in the [directory structure](https://github.com/keep-network/keep-common) section of `keep-common` README.

Done in https://github.com/keep-network/keep-common/commit/7597045701348109e0b84a2dbd59545666c9ba99

pkg/chain/ethlike/generator/template/resubscribe.go.tmpl
```diff
@@ -0,0 +1,59 @@
+// Code generated - DO NOT EDIT.
```
pdyraga, line 1: We were placing all of the templates in `tools/generators` so far, now we are breaking this layout. It is unclear to me if this is something we have reasons for? (cc @Shadowfiend)

pkg/chain/celo/celoutil/gen/error_resolver.go
```diff
@@ -1,22 +1,18 @@
-package celoutil
+// Code generated - DO NOT EDIT.
+
+package gen
```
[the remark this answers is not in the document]
lukasz-zimnoch: I think this will be inconvenient in places where we'll also use the original `celo/celoutil` package. Also, that package structure looks weird and adds a redundant level to the package structure. Personally, I prefer to have everything in `celo/celoutil` package instead of trying to separate generated code at all costs by complicating the package structure and playing with import aliases in the imported code. After all, generated files are easily distinguishable because they have `Code generated - DO NOT EDIT.` comments at their top.

pkg/chain/celo/celoutil/gen/error_resolver_test.go
```diff
@@ -1,15 +1,14 @@
-package celoutil_test
+package gen_test
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is something that has been inherited from Ethereum's error resolver tests. I think we can safely change it to just `package gen`. Regarding the second point, I'm also for keeping tests in the same package as tested code. If we separate them I think we can expect some complications and we may be forced to export more things than we should.

pkg/chain/ethlike/generator/template/resubscribe.go.tmpl
lukasz-zimnoch: The main motivation for placing those files in `pkg/chain/ethlike/generator` was the fact they are very specialized ETH-like tools that are used only by ETH-like chain implementations. They are not as generic and ubiquitous as tools placed in `tools/generators`. This is why I felt right to keep them in the `ethlike` package close to other tools.

pkg/chain/celo/celoutil/gen/error_resolver.go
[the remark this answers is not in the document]
pdyraga: `gen` is just a pattern that indicates we are dealing with generated code. I do agree import aliases could be unwieldy but at the same time, I do not think we should use `gen` as a package name. I need to think more about it, maybe there is a name under which we can aggregate all of them: `celo/ceolutil/gen/somename` 🤔

pkg/chain/ethlike/generator/template/resubscribe.go.tmpl
Shadowfiend: Hello, I'm super-late here! `tools/` is meant for executable tools. Something in `tools/` can use library code in `pkg/` if needed, but if we're talking about a code generation tool, it belongs in `tools/`. We should never `go run` something in `pkg/`.

pdyraga:
In another galaxy, 8 months later...

We see that Celo and Ethereum are diverging. One example is https://github.com/keep-network/keep-common/pull/84 where Ethereum changed the transaction pricing mechanism and this was not propagated to Celo. Also, Celo plans to add a gateway fee to transaction parameters that will not be available in Ethereum.

I think holding off with the work on this PR was a good choice and I vote for closing it without merging. What do you say @lukasz-zimnoch?

lukasz-zimnoch:
> In another galaxy, 8 months later...
> 
> We see that Celo and Ethereum are diverging. One example is #84 where Ethereum changed the transaction pricing mechanism and this was not propagated to Celo. Also, Celo plans to add a gateway fee to transaction parameters that will not be available in Ethereum.
> 
> I think holding off with the work on this PR was a good choice and I vote for closing it without merging. What do you say @lukasz-zimnoch?

Yep, I agree!

PR #70, opened 2021-03-04
Refactor `ContractAddress` method in config structs

lukasz-zimnoch:
`ContractAddress` method in both Ethereum's and Celo's config structs should return a plain `common.Address` type instead of a pointer. This should eliminate the need for nil-checks in the client code.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ replace github.com/urfave/cli => github.com/keep-network/cli v1.20.0
 require (
 	github.com/allegro/bigcache v1.2.1 // indirect
 	github.com/aristanetworks/goarista v0.0.0-20190924011532-60b7b74727fd // indirect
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -7,7 +7,6 @@ replace github.com/urfave/cli => github.com/keep-network/cli v1.20.0
 require (
 	github.com/allegro/bigcache v1.2.1 // indirect
 	github.com/aristanetworks/goarista v0.0.0-20190924011532-60b7b74727fd // indirect
-	github.com/btcsuite/btcd v0.20.1-beta // indirect
```
pdyraga, line 10: What is happening here? Seems like an unrelated change.

lukasz-zimnoch: `go mod tidy` cleaned it up.

pdyraga (approved):

PR #71, opened 2021-03-10
Adjust Ethereum contracts to use EIP155 signer

Shadowfiend:
go-ethereum 1.10.0 has switched to denying non-EIP155 signatures by default.
Although our code uses go-ethereum's own ABI binding generation, the version
currently in our dependencies never added support for EIP155 signers, and
instead always uses Homestead signers for the transactors it provides helpers
for.

In the short term, this commit manually creates the transactor options for the
account key, forcing an EIP155 signer when doing signing. When the go-ethereum
dependency gets bumped past 1.9.25, we will be able to move back to
`NewKeyedTransactorWithChainID`, which uses an EIP155 signer under the hood.

### DIFF
--- a/tools/generators/ethereum/command.go.tmpl
+++ b/tools/generators/ethereum/command.go.tmpl
@@ -204,6 +204,14 @@ func initialize{{.Class}}(c *cli.Context) (*contract.{{.Class}}, error) {
         return nil, fmt.Errorf("error connecting to Ethereum node: [%v]", err)
     }
 
+   	chainID, err := client.ChainID(context.Background())
+	if err != nil {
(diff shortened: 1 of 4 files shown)


Shadowfiend:
This PR is open as a draft because it is a reference point for the 1.7.0 release of the keep-ecdsa client that includes only this commit. Separate PR incoming off of master.

pdyraga:
We properly tagged 34905d2 as `v1.4.0` and released it. We can now undraft this PR and treat it as a base for `v1.4.1`.

pdyraga (approved):

PR #72, opened 2021-03-10
Backport: Adjust Ethereum contracts to use EIP155 signer

Shadowfiend:
go-ethereum 1.10.0 has switched to denying non-EIP155 signatures by default.
Although our code uses go-ethereum's own ABI binding generation, the version
currently in our dependencies never added support for EIP155 signers, and
instead always uses Homestead signers for the transactors it provides helpers
for.

In the short term, this commit manually creates the transactor options for the
account key, forcing an EIP155 signer when doing signing. When the go-ethereum
dependency gets bumped past 1.9.25, we will be able to move back to
`NewKeyedTransactorWithChainID`, which uses an EIP155 signer under the hood.

### DIFF
--- a/tools/generators/ethlike/contract.go.tmpl
+++ b/tools/generators/ethlike/contract.go.tmpl
@@ -61,6 +61,9 @@ func New{{.Class}}(
 	// FIXME bumps beyond 1.9.25.
 	key := accountKey.PrivateKey
 	keyAddress := crypto.PubkeyToAddress(key.PublicKey)
+	if chainId == nil {
+		return nil, fmt.Errorf("no chain id specified")
+	}
 	transactorOptions := &bind.TransactOpts{
 		From: keyAddress,
 		Signer: func(_ types.Signer, address common.Address, tx *types.Transaction) (*types.Transaction, error) {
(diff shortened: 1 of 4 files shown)


Shadowfiend:
I'm fairly certain this works as expected for Celo, but not 100% certain.

tools/generators/ethlike/contract.go.tmpl
```diff
@@ -44,6 +45,7 @@ type {{.Class}} struct {
 
 func New{{.Class}}(
     contractAddress common.Address,
+    chainId *big.Int,
```
pdyraga, line 48: We don't do it for the hotfix but I think here we should add validation for `chainId` to error if it is null. EIP155 signer does not do the validation, and if chain ID is nil, the transaction is rejected as not properly signed but without any information that the lack of chain ID is the case.

pdyraga: Note that the original `go-ethereum` code is performing this check before constructing EIP155 signer: https://github.com/ethereum/go-ethereum/commit/ed0670cb17a96aafeb9eaaeb9765a42fb6bb5663#diff-18aeac66ddb948fefafc8c96289984cd90c94c54d0616e96165ffff8865b6250R143

pdyraga: Done in https://github.com/keep-network/keep-common/pull/72/commits/94d25202ee9d1a1a6c66e6df72b24f49ba6b046a.

pdyraga:
Tested with Celo using the code from https://github.com/keep-network/keep-ecdsa/pull/726. Everything works as expected. I am ready to merge it but I'll need @Shadowfiend's approval on my change.

pdyraga (approved):

nkuba (approved):

PR #74, opened 2021-03-26
Cleanup of Go workflow

michalinacienciala:
This PR introduces a couple of cleanup changes in Go workflow, as part
of work on RFC-18. Those changes are:
* Moving `scan` job from `gosec.yml` to `client.yml` (there is no need
to keep them in separate workflows)
* Adding `client-` prefix to `scan` and `build-and-test` jobs (to avoid
confusion in the future with other similar jobs that we may want to add
or remove from required list in branch protection settings)
* Removing test result puvblishing step (we've decided to resign from
using `EnricoMi/publish-unit-test-result-action` action, as it worked
in a confising way sometimes - was assigning the tests results to
incorrect workflows)
* Removing unnecessary comment


michalinacienciala:
nkuba, following jobs can be configured as required in GitHub's settings:
* `client-build-and-test`
* `client-scan`

.github/workflows/client.yml
```diff
@@ -25,20 +24,18 @@ jobs:
...
         run: go get gotest.tools/gotestsum
 
       - name: Run Go tests
-        run: gotestsum --junitfile test-results/unit-tests.xml
+        run: gotestsum
 
-      - name: Publish unit test results
-        uses: EnricoMi/publish-unit-test-result-action@v1
-        if: always()
+  client-scan:
```
nkuba, line 33: I'm wondering which name would be better: `client-scan` or `client-gosec`. No strong opinion, so merging it as is. We can always update it in the future.

nkuba (approved):
Maybe

PR #75, opened 2021-04-19
Add Go linting to the client.yml workflow

michalinacienciala:
As part of work on RFC-18 a job linting the Go code has been added to
`client.yml` workflow. Action for linting of Go code has been also added
(as separate repository) as part of this work.

Running the worklow with linting revealed a number of issues found
by the linter ([see here](https://github.com/keep-network/keep-common/runs/2379469245?check_suite_focus=true)).
All of them are also fixed as part of this PR.

@nkuba, after merging this PR into master please configure the
`client-lint` job in GitHub's settings as required for PR merging.

### DIFF
--- a/pkg/chain/ethlike/block_counter.go
+++ b/pkg/chain/ethlike/block_counter.go
@@ -8,6 +8,7 @@ import (
 	"time"
 )
 
+// BlockCounter represents a block counter
 type BlockCounter struct {
 	structMutex         sync.Mutex
 	latestBlockHeight   uint64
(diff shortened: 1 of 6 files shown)


pkg/chain/ethlike/block_counter.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 )
 
+// BlockCounter represents a block counter
```
nkuba, line 11: We need to add dots at the ends of all these comments we're adding in this PR.

nkuba (changes requested):

pkg/chain/ethlike/block_counter.go
michalinacienciala: Done in f1c11be35fce9272ea39e30ad61d1ee28b03ec80.

nkuba (approved):

PR #77, opened 2021-05-05
Expose checkStoragePermission as a Public Function

beaushinkle:
Having this be public would be broadly useful. See: https://github.com/keep-network/keep-ecdsa/pull/756/files#diff-ca2057cf86a8e1934767fc81dc8dfeb18e843c2bd3d52d99442a248553b36296R26

Tagging @nkuba for review

### DIFF
--- a/pkg/persistence/disk_persistence.go
+++ b/pkg/persistence/disk_persistence.go
@@ -18,7 +18,7 @@ const (
 
 // NewDiskHandle creates on-disk data persistence handle
 func NewDiskHandle(path string) (Handle, error) {
-	err := checkStoragePermission(path)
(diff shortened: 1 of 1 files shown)


Shadowfiend:
Had some light discussion with @beaushinkle on what we really want here, which is to adjust `persistence.Handle` to provide support for multiple storage areas, some of which need the current/archive/snapshot behavior and some of which just store data normally. For now, we can expose functions like this one, but ideally we'd export a higher-level abstraction and leave these low-level details unexported.

We can explore this further in the future.

Shadowfiend (approved):
This is pretty constrained in blast radius, and tests are passing. Let's roll it in.

PR #82, opened 2021-09-01
Balance monitoring retries

nkuba:
We want to use a retry mechanism for single executions of the balance check.
The retries will be performed during the period of `retryTimeout`, logging a warning on each error. When the timeout is hit an error will be logged and retries stopped. The next balance check will be triggered at the next tick.

### DIFF
--- a/pkg/chain/ethlike/balance_monitor.go
+++ b/pkg/chain/ethlike/balance_monitor.go
@@ -63,12 +63,21 @@ func (bm *BalanceMonitor) Observe(
 		ticker := time.NewTicker(tick)
 		defer ticker.Stop()
 
+		checkBalance := func() {
+			err := wrappers.DoWithDefaultRetry(retryTimeout, check)
+			if err != nil {
+				logger.Errorf("balance monitor error: [%v]", err)
+			}
+		}
+
+		// Initial balance check at monitoring start.
+		checkBalance()
+
 		for {
 			select {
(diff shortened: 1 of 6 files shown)


pkg/chain/ethlike/balance_monitor.go
```diff
@@ -22,17 +25,26 @@ func NewBalanceMonitor(balanceSource BalanceSource) *BalanceMonitor {
 // Observe starts a process which checks the address balance with the given
 // tick and triggers an alert in case the balance falls below the
 // alert threshold value.
+// The balance check will be retried in case of an error up to the retry timeout.
 func (bm *BalanceMonitor) Observe(
 	ctx context.Context,
 	address Address,
 	alertThreshold *Token,
 	tick time.Duration,
+	retryTimeout time.Duration,
```
pdyraga, line 34: Non-blocking but shouldn't we validate `retryTimeout < tick` so that we do not have two checks overlapping?

pdyraga (approved):

pkg/chain/ethlike/balance_monitor.go
nkuba: Yeah, good idea.

nkuba: On the other hand, don't we want to leave it to the implementation using this funcion?

PR #85, opened 2021-10-12
EIP-1559 Ethereum transaction options

pdyraga:
Ethereum mining waiter has been already updated to support dynamic fee
(EIP-1559) transactions but ethutil.TransactionOpts still exposed only
GasPrice option. Setting this option resulted in marking the outgoing
transaction as a legacy type transaction. Given that the client should
use dynamic fee transactions, GasPrice option has been replaced with
GasFeeCap and GasTipCap. This change is only for Ethereum. Celo still
supports GasPrice as EIP-1559 affects only Ethereum.

### DIFF
--- a/pkg/chain/ethereum/ethutil/transaction_opts.go
+++ b/pkg/chain/ethereum/ethutil/transaction_opts.go
@@ -9,23 +9,32 @@ import (
 // TransactionOptions represents custom transaction options which can be used
 // while invoking contracts methods.
 type TransactionOptions struct {
-	// GasLimit specifies a gas limit to set on a transaction call; should be
-	// ignored if set to 0.
+	// GasLimit specifies a gas limit to set for a transaction; estimated if
+	// set to 0.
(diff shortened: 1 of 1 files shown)


pkg/chain/ethereum/ethutil/transaction_opts.go
```diff
@@ -9,23 +9,32 @@ import (
 // TransactionOptions represents custom transaction options which can be used
 // while invoking contracts methods.
 type TransactionOptions struct {
-	// GasLimit specifies a gas limit to set on a transaction call; should be
-	// ignored if set to 0.
+	// GasLimit specifies a gas limit to set for a transaction; estimated if
+	// set to nil.
```
lukasz-zimnoch, line 13: `set to 0`?

pdyraga: Yes, I'll fix!

PR #91, opened 2022-07-13
Remove celo

nkuba:
In this PR we remove code related to Celo as it's no longer needed in V2.


pdyraga:
Can we also move everything from `ethlike` to `ethereum`?

pdyraga:
> Can we also move everything from `ethlike` to `ethereum`?

Not in this PR obviously, this should go separate as it will create some conflicts.

pdyraga (approved):

nkuba:
> > Can we also move everything from `ethlike` to `ethereum`?
> 
> Not in this PR obviously, this should go separate as it will create some conflicts.

I like the idea!

PR #92, opened 2022-07-13
Move ethlike to ethereum

nkuba:
This PR is a follow-up from https://github.com/keep-network/keep-common/pull/91#issuecomment-1183354389

We no longer want to support Ethereum like (Celo) chain, hence we moved `ethlike` to `ethereum`.


pkg/chain/ethereum/ethutil/adapter.go
```diff
@@ -2,36 +2,38 @@ package ethutil
```
pdyraga, line 2: I can't find the reasoning in commit messages but I am pretty sure the reason for introducing the concept of an adapter was to "flatten" types returned by go-ethereum functions into common ones so that we can use the same API for Celo and Ethereum.

Not in this PR and just a note for future that we should consider getting rid of it, cc @lukasz-zimnoch

pdyraga (approved):

PR #98, opened 2022-08-15
Replace urfave/cli with spf13/cobra in commands generator

nkuba:
In the keep-client, we replaced `urfave/cli` with `spf13/cobra`:
- `start` command: https://github.com/keep-network/keep-core/pull/3090
- `ping` command: https://github.com/keep-network/keep-core/pull/3159

We want to make `ethereum` command work with the `cobra`. For that we need to refactor `keep-common` commands generator for contract interactions.

The changes were used in keep-core in https://github.com/keep-network/keep-core/pull/3181


pkg/chain/ethereum/wei_test.go
```diff
@@ -84,7 +84,7 @@ func TestUnmarshalTextValue(t *testing.T) {
 			expectedError: fmt.Errorf("failed to parse value: [3 wei ]"),
 		},
```
tomaszslabon, line 86: The blank line is probably not needed.

pkg/cmd/cmd.go
```diff
@@ -4,16 +4,17 @@ package cmd
...
 )
 
 const (
-	blockFlag        string = "block"
-	blockShort       string = "b"
-	transactionFlag  string = "transaction"
-	transactionShort string = "t"
+	blockFlag  string = "block"
+	blockShort string = "b"
 	// SubmitFlag allows for urfave/cli definition and lookup of a boolean
```
tomaszslabon, line 18: Since we removed the `urfave/cli` package, this command should be updated, so that it does not mention it.

pkg/chain/ethereum/wei_test.go
nkuba: 98cf58736765ea8446021bd24371789574580879

pkg/cmd/cmd.go
nkuba: fd26c8128e2c05b4ef7a406e5d161eed7f0f77e5

tomaszslabon (approved):

PR #99, opened 2022-08-25
Configure Ethereum network as enum in the etherem config

nkuba:
We add Network enumeration of supported Ethereum Networks like Mainnet
and Goerli. We include a property specifing the network in the
ethereum.Config struct.

### DIFF
--- a/pkg/chain/ethereum/network.go
+++ b/pkg/chain/ethereum/network.go
@@ -1,0 +1,24 @@
+package ethereum
+
+// Network is a type used for ethereum networks enumeration.
+type Network int
+
+// Ethereum networks enumeration.
+const (
+	Unknown Network = iota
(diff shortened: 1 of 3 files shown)


pkg/chain/ethereum/network.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Ethereum networks enumeration.
+const (
+	Unknown Network = iota
+	Mainnet
+	Goerli
+	Developer
+)
+
+func (n Network) String() string {
```
dimpar, line 14:
```
func (n Network) String() string {
    return [...]string{"unknown", "mainnet", "goerli", "developer"}[n]
}
```
up to you..

dimpar (approved):

PR #103, opened 2022-09-12
Add Delete function to persistent storage

nkuba:
Expose a Delete function that can be used to remove file from persistent sotarge. Add disk storage implementation.

### DIFF
--- a/pkg/persistence/disk_persistence.go
+++ b/pkg/persistence/disk_persistence.go
@@ -40,7 +40,7 @@ func NewDiskHandle(path string) (Handle, error) {
 	}
 
 	snapshotSuffixGenerator := func() string {
-		timestamp := time.Now().UnixNano() / int64(time.Millisecond)
+		timestamp := time.Now().UnixMilli()
 		return fmt.Sprintf(".%d", timestamp)
 	}
 
--- a/pkg/persistence/disk_persistence.go
+++ b/pkg/persistence/disk_persistence.go
@@ -80,7 +80,7 @@ func (ds *diskPersistence) Save(data []byte, dirName, fileName string) error {
 		return err
 	}
 
-	return Write(fmt.Sprintf("%s/%s/%s", dirPath, dirName, fileName), data)
+	return Write(filepath.Join(dirPath, dirName, fileName), data)
 }
 
 func (ds *diskPersistence) Snapshot(data []byte, dirName, fileName string) error {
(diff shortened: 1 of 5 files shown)


pkg/persistence/disk_persistence.go
```diff
@@ -149,6 +149,13 @@ func (ds *diskPersistence) Archive(directory string) error {
 	return moveAll(from, to)
 }
 
+func (ds *diskPersistence) Delete(dirName string, fileName string) error {
+	dirPath := ds.getStorageCurrentDirPath()
+	filePath := filepath.Join(dirPath, dirName, fileName)
```
pdyraga, line 154: Can we take the opportunity and update all `fmt.Sprintf("%s/%s/%s",...)` occurrences in this file?

pkg/persistence/disk_persistence_test.go
```diff
@@ -499,3 +499,55 @@ func TestDiskPersistence_AppendToArchive(t *testing.T) {
...
+
+	bytesToTest := []byte{115, 111, 109, 101, 10}
+
+	diskPersistence.Save(bytesToTest, dirName1, fileName11)
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		if err != nil {
+			t.Fatalf("Dir [%+v] was supposed to be created", path)
+		}
+	}
```
pdyraga, line 515: Shouldn't we instead check if the file was created?

Something like:
```
	pathToFile := fmt.Sprintf("%s/%s/%s", pathToCurrent, dirName1, fileName11)

	if _, err := os.Stat(pathToFile); os.IsNotExist(err) {
		t.Fatalf("file [%+v] was supposed to be created", pathToFile)
	}
```

pkg/persistence/disk_persistence_test.go
```diff
@@ -499,3 +499,55 @@ func TestDiskPersistence_AppendToArchive(t *testing.T) {
...
+	err := diskPersistence.Delete(dirName1, fileName11)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	if _, err := os.Stat(path); !os.IsNotExist(err) {
+		if err != nil {
+			t.Fatalf("Dir [%+v] was supposed to be moved", path)
+		}
+	}
```
pdyraga, line 526: I am confused. I would expect the file to be deleted, not the directory. See that `path` points to the directory not to the file. Also s/moved/deleted.

pkg/persistence/disk_persistence.go
nkuba: 41ca4fdc902a4c0b4c13d0007d811859bf9c9ba0

pkg/persistence/disk_persistence_test.go
nkuba: The tests were confusing, I updated them in 286069a18defe0155fc24fd7b047eff5cafe4a2f and 22a0f19484663ed9a39425df79aa9c09ad154798.

pkg/persistence/disk_persistence_test.go
nkuba: 💯 see https://github.com/keep-network/keep-common/pull/103#discussion_r969445720

pdyraga (approved):

PR #104, opened 2022-09-15
Separate persistences for key store and generated parameters

nkuba:
We want to have two types of persistence handles that store data on
disk:
- for sensitive data (e.g. keys) we don't want any data to be removed,
  but archived instead, we also need snapshots, this requires a specific
  directory structure with `current`, `archive` and `snapshot`
  subdirectories
- for regular data (e.g. generated parameters) we want to store and
  delete data in a simple directory structure, without `current`
  subdirectory.

Here we separate the interfaces for persistence handles:
- RWHandle - common interface for reading and saving data
- BasicHandle - simple interface for the regular data to save, read and
  remove them
- ProtectedHandle - for sensitive data (i.e. keystore) to archive and
  snapshot instead of delete


.vscode/configurationCache.log
```diff
@@ -0,0 +1 @@
+{"buildTargets":[],"launchTargets":[],"customConfigurationProvider":{"workspaceBrowse":{"browsePath":[],"compilerArgs":[]},"fileIndex":[]}}
```
lukasz-zimnoch, line 1: Is the `.vscode` directory added on purpose?

nkuba: Nope!

lukasz-zimnoch (approved):

PR #105, opened 2022-09-26
Merge diagnostics and metrics into a single clientinfo registry

pdyraga:
Previously, we were exposing diagnostics and metrics on separate ports but the reality was that we had two HTTP servers exposing exactly the same content: the diagnostics port could be queried for both /diagnostics and /metrics and the metrics port could be queried for both /metrics and /diagnostics. This change simplifies the setup into a single clientinfo package and registry exposing metrics and diagnostics.

The clientinfo package name was inspired by https://pkg.go.dev/debug/buildinfo

TODO:
- [x] Integrate it with keep client
- [x] Should we rename `Observer` to `MetricsObserver`?


pkg/clientinfo/observer.go
```diff
@@ -1,4 +1,4 @@
-package metrics
+package clientinfo
```
nkuba, line 1: > Should we rename Observer to MetricsObserver?

We could do that.

nkuba (approved):
I let myself commit unit tests and a fix to this branch.

I tested the changes with Keep Client and Prometheus Service Discovery. It works as expected.

pkg/clientinfo/observer.go
pdyraga: [86dacb6](https://github.com/keep-network/keep-common/pull/105/commits/86dacb6a965501d8ba579361d0cf01941dc62b7c)

nkuba (approved):

PR #109, opened 2022-12-02
Require parent context in wrappers

nkuba:
Instead of using `context.Background()` context in the wrappers we require a parent context to be passed so the retries and waits are correctly aborted when the parent context is done.


nkuba:
This change was already integrated in keep-core: https://github.com/keep-network/keep-core/pull/3406/commits/70bba327634895de807a903e30875fe77a02c3ca

pdyraga (approved):

PR #116, opened 2023-05-01
Support bytes20 and bytes32 in Ethereum contract bindings

pdyraga:
There are quite a few functions in our Solidity contracts that accept `bytes20` or `bytes32` as a parameter. One of many examples are [`requestHeartbeat(bytes20 walletPubKeyHash, bytes calldata message)`](https://github.com/keep-network/tbtc-v2/blob/eddb2b5f9468daf18f27c45c06b70ab05bda78ef/solidity/contracts/bridge/WalletCoordinator.sol#L352) from the `WalletCoordinator` contract and [`getWallet(bytes32 walletID)`](https://github.com/keep-network/keep-core/blob/6302a42470772762b63cf6f584878aa5d4d4d2d2/solidity/ecdsa/contracts/WalletRegistry.sol#L1054) from the `WalletRegistry` contract. Given there was no parsing function defined in our Ethereum contract parsing logic, all functions accepting `bytes32` and `bytes20` were skipped during the code generation:

```
WARNING: Unsupported param type for method getWallet:
  ABI Type: bytes32
  Go Type:  [32]byte
  the method won't be callable with 'ethereum' command
```

This changeset adds two parsing functions: one for `bytes20` and another for `bytes32` Solidity type. In theory, we could have other variations of `bytesN` but this is not the case for our codebase, and defining a generic function is more complicated given we do not deal with slices but fixed-size arrays. Even though the proposed solution does not work for all `bytesN` types, it is enough for now.

## Testing

Testing this change is easy.

First, see how it works without this change. Go to your `keep-network/keep-core` repository copy and execute `make all`. You will see a bunch of warnings like this:

```
WARNING: Unsupported param type for method getWallet:
  ABI Type: bytes32
  Go Type:  [32]byte
  the method won't be callable with 'ethereum' command
WARNING: Unsupported param type for method isWalletRegistered:
  ABI Type: bytes32
  Go Type:  [32]byte
  the method won't be callable with 'ethereum' command
WARNING: Unsupported param type for method isWalletMember:
  ABI Type: bytes32
  Go Type:  [32]byte
  the method won't be callable with 'ethereum' command
```

Check out this branch locally, go to `keep-network/keep-core` repository, and add replace directive in `go.mod` to point to the checked-out code. In my case, it is `github.com/keep-network/keep-common => /Users/piotr/git/keep-network/keep-common`. Then run `make all` again.

All warnings should be gone and you should see the code generated for the previously skipped functions.

I did this testing locally and I was able to call some functions with the generated code. Example:
```
$ ./keep-client ethereum tbtc wallet-coordinator request-heartbeat \
0x3805eed0bb0792eff8815addedb36add2c7257e5 \
0xFFFFFFFFFFFFFFFF0000000000000000 \
--config configs/config.deployer.toml --submit
```

### DIFF
--- a/pkg/utils/decode/bytes.go
+++ b/pkg/utils/decode/bytes.go
@@ -6,6 +6,8 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
+// ParseBytes20 parses `string` into `[20]byte` type. The input string must have
+// 20 bytes hex-encoded.
 func ParseBytes20(str string) ([20]byte, error) {
 	bytesArray := [20]byte{}
 	slice, err := hexutil.Decode(str)
@@ -20,6 +22,8 @@ func ParseBytes20(str string) ([20]byte, error) {
 	return bytesArray, nil
 }
 
+// ParseBytes32 parses `string` into `[32]byte` type. The input string must have
+// 32 bytes hex-encoded.
 func ParseBytes32(str string) ([32]byte, error) {
 	bytesArray := [32]byte{}
 	slice, err := hexutil.Decode(str)
--- a/pkg/utils/decode/bytes.go
+++ b/pkg/utils/decode/bytes.go
@@ -1,0 +1,35 @@
+package decode
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+func ParseBytes20(str string) ([20]byte, error) {
+	bytesArray := [20]byte{}
+	slice, err := hexutil.Decode(str)
+	if err != nil {
+		return bytesArray, err
+	}
+	if len(slice) != 20 {
+		return bytesArray, fmt.Errorf("expected 20 bytes array; has: [%v]", len(slice))
+	}
+
+	copy(bytesArray[:], slice)
+	return bytesArray, nil
+}
+
+func ParseBytes32(str string) ([32]byte, error) {
+	bytesArray := [32]byte{}
+	slice, err := hexutil.Decode(str)
+	if err != nil {
+		return bytesArray, err
+	}
+	if len(slice) != 32 {
+		return bytesArray, fmt.Errorf("expected 32 bytes array; has: [%v]", len(slice))
+	}
+
+	copy(bytesArray[:], slice)
+	return bytesArray, nil
+}
--- a/pkg/utils/decode/bytes_test.go
+++ b/pkg/utils/decode/bytes_test.go
@@ -1,0 +1,126 @@
+package decode
+
+import (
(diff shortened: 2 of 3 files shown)


pkg/utils/decode/bytes.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+// ParseBytes20 parses `string` into `[20]byte` type. The input string must have
+// 20 bytes hex-encoded.
+func ParseBytes20(str string) ([20]byte, error) {
+	bytesArray := [20]byte{}
+	slice, err := hexutil.Decode(str)
```
lukasz-zimnoch, line 13: Why not just `hex.DecodeString` from the standard library? Is there any particular reason to use `hexutil.Decode` here?

pkg/utils/decode/bytes.go
```diff
@@ -0,0 +1,39 @@
...
+
+	copy(bytesArray[:], slice)
+	return bytesArray, nil
+}
+
+// ParseBytes32 parses `string` into `[32]byte` type. The input string must have
+// 32 bytes hex-encoded.
+func ParseBytes32(str string) ([32]byte, error) {
+	bytesArray := [32]byte{}
+	slice, err := hexutil.Decode(str)
```
lukasz-zimnoch, line 29: Same comment about `hex.DecodeString`

pkg/utils/decode/bytes.go
pdyraga: `hexutil` package is already used to [decode `[]byte`](https://github.com/keep-network/keep-common/blob/2632871d696cd604823591142e2e1f7b61c96b60/tools/generators/ethereum/contract_parsing.go#L233) and to [decode `*big.Int`](https://github.com/keep-network/keep-common/blob/2632871d696cd604823591142e2e1f7b61c96b60/tools/generators/ethereum/contract_parsing.go#L237). It also works with `0x` prefix while `hex.DecodeString` from the standard library expects the prefix to be skipped.

I was thinking about using the standard library version here but then I'd have to deal with prefix, and clean up `contract_parsing.go` by introducing another function in `bytes.go` that would deal with `[]byte` for consistency. Then, I'd still have the `*big.Int` parsing left.

I know it is not ideal but it's simpler. And I think we are bound to `hexutil` anyway so we should either use it consistently or replace other usages as well.

lukasz-zimnoch: Ok, that's clear now. Thanks!

lukasz-zimnoch (approved):

